PR #21, opened 2025-03-19
Feat/testing

GondekNP:
Testing scaffold set up and some basic testing enabled for Scalars (for now, will probably merge and continue updating this branch for new Classes)

- As of now, legit looking errors for dimensional analysis logic, but nothing we can't handle!


GondekNP:
This will be subsumed by `feat/distributions`, which added a logic to `EngineValue`, `Scalar`s and `Distrubution`s with associated testing

PR #22, opened 2025-03-20
Reorganize examples and add asserts

sampottinger:
Prepare to start interpreter implementation by:

 - Organizing examples so that there are feature-level examples vs full simulation examples.
 - Adding asserts to full simulation examples.

This also adds the idea of an assertion to the language.


GondekNP:
Sam and I discussed this in person and determined that `assertions` might be the best way for end users to write little test cases / sanity checks. This would be a good opportunity to get feedback from @mzomer and @lucialayr as they work thru their initial toy demo

GondekNP:
@lucialayr and @mzomer - we'll merge this now but let's revisit later on after we all have some more familiarity writing simulations

PR #23, opened 2025-03-20
Migrate style for constructor.

sampottinger:
Moving from using `new` prefix on parameter names for constructors to avoid shadowing to instead expect parameter names to match instance fields and using `this` to work around shadowing. However, use of `this` outside of the constructor to access instance variables remains discouraged.


GondekNP (approved):
**THIS** is my favorite merge request yet. I can't wait to keep working on **THIS** repo, wow what a wonderful day **THIS** has been.

LGTM - to the future java historians, I am the cause of **THIS** refactor :)

PR #25, opened 2025-03-20
Add WASM build

sampottinger:
Add support for building a WASM distribution via TeaVM. This uses a reduced build size without a CLI for embedding into web.

Sorry for the runaround on this @GondekNP. I determined that the WASM build was kinda large and took a while to build because [picocli](https://picocli.info/) has some options to parse a whole bunch of very specific things like `InetAddress` and the like. This ends up bringing a lot more from the JDK than is actually required. Creating a [facade](https://refactoring.guru/design-patterns/facade) for the library and pointing TeaVM to that instead of the CLI entrypoint cuts out those unneeded standard classes and keeps the build much smaller. I think this is the right move since Picocli does all of that for convienence since it expects a full JDK whereas TeaVM does tree shaking to slim down the WASM artifact.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -66,6 +66,28 @@ jobs:
         with:
           name: fatJar
           path: build/libs/joshsim-fat.jar
+  buildWasm:
+    needs: buildJava
+    environment: Build
+    runs-on: ubuntu-latest
+    name: Build WASM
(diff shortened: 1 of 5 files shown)


GondekNP:
Nice work Sam, looks like just what we need for browser deployment :cactus:

PR #26, opened 2025-03-20
Add the bridge between the engine and interpreter.

sampottinger:
Added the bridge that acts as the interface between the interpreter in the form of the ANTLR visitor and the engine. I think I am ok with engine types going into the interpreter side to avoid too much type duplication but I do want to formalize an object that stands between the two so that it is easier to decouple behavioral changes on both ends. There are a few things that I am hoping to have in separate PRs after this is in:

 - Converter implementation on engine side: I have some ideas on how to implement this potentially with DirectConverter and TransitiveConverter. I didn’t include on this merge request though since it felt like that was more engine-insular.
 - VariableResolver on the language side: I don’t want to optimize too prematurely but scope resolution may be slow. I think there’s an easy answer here on the language side to cache the results of scope path resolution but I didn’t want to make this merge request too big.
 - QueryResolver on language side: I don’t want to optimize too prematurely but I think there’s an easy move here on the language side to check if the user is passing a constant and can cache the results of a spatial query (CachingQueryResolver) as I think that is the slowest part of the system right now. That may only be possible on the language side though where I can check that it is a constant. It does involve adding a key to patches though so I might propose a merge request to get feedback.

If this and those three merge requests are in, I think the ANTLR visitor is pretty straightforward / thin.


sampottinger:
Blocked on #27.

sampottinger:
Note that this did require adding EntityValue. I think this is desired though as the language would allow one to make and work with distributions of Entities or save Entities to variables. Those operations without this I think could involve a lot of inspection so I would prefer to have EntityValue but would love to get your thoughts @GondekNP !

src/main/java/org/joshsim/engine/func/EntityScope.java
```diff
@@ -0,0 +1,69 @@
...
+    return expectedAttrs;
+  }
+
+  /**
+   * Extract all attribute names from an entity's event handlers.
+   *
+   * @param target the Entity from which to extract attribute names.
+   * @return Set of attribute names found in the entity's event handlers.
+   */
+  private Set<String> getAttributes(Entity target) {
```
GondekNP, line 63: This is more of a java optimization thing, but, is there any situation where we have one of these map operations and _don't_ want to just use StreamAPI? I figured intuitively there might be some overhead for instantiating the stream vs simply iterating, but perhaps the move is to always always go for it

src/main/java/org/joshsim/engine/simulation/Query.java
```diff
@@ -18,15 +18,15 @@
  */
 public class Query {
 
-  private final int step;
+  private final long step;
```
GondekNP, line 21: I know it is for consistency, but fun to thing about the future futures we are predicting where we need a long over an int :stuck_out_tongue_closed_eyes: - heat death of the universe

src/main/java/org/joshsim/engine/value/EngineValueFactory.java
```diff
@@ -79,6 +80,16 @@ public EngineValue build(BigDecimal innerValue, Units units) {
...
   }
 
+  /**
+   * Build a new EngineValue from an Entity.
+   *
+   * @param entity the value to decorate in an EngineValue.
+   * @returns decorated version of entity.
+   */
+  public EngineValue build(Entity entity) {
+    return new EntityValue(caster, entity);
```
GondekNP, line 90: How do we expect `EntityValue` to behave differently than `EngineValue`?

GondekNP (commented):
Looks great Sam - only comment is for us to check in about `EntityValue`

src/main/java/org/joshsim/engine/value/EngineValueFactory.java
GondekNP: (basically we want a much more restricted set of `EngineValue` logic, minus all the arithmetic operations - so this is as complicated as `EntityValue` will get)

PR #29, opened 2025-03-20
Feat/distributions abstraction

sampottinger:


GondekNP:
- Plan to move innerType / rootType logic into EngineValue rather than EngineValueTuple

PR #31, opened 2025-03-22
Geometry Implementation (Using Spatial4j Shapes)

GondekNP:
- Basic `Geometry` Class (wrapper for Spatial4j with methods for interects and distance)
- `GeometryFactory` for Geometry flavors - for now just Squares for cell patches (will be extended for `Circle`)
- `Grid` for containing cells, and `GridBuilder` for creating `Grids`

Some tweaks made to interfaces within `Entity` as well, but nothing major. Biggest change is that we have a generic `Geometry` class rather than `Circle`, `Rect`, etc etc, primarily because `Geometry.shape` is doing the work and we don't really have any differing implementations other than what is currently handled by Spatial4j.


sampottinger:
Thanks @GondekNP! I’ll provide a review in the morning.

src/main/java/org/joshsim/engine/entity/Disturbance.java
```diff
@@ -6,11 +6,54 @@
...
  */
-public interface Disturbance extends SpatialEntity {}
+public class Disturbance extends SpatialEntity {
+
+  /**
+   * Constructs a disturbance entity with the given geometry.
+   *
+   * @param geometry the geometry of the disturbance
+   */
+  public Disturbance(Geometry geometry) {
```
sampottinger, line 26: For a future merge request, might be opportunity to move this up into SpatialEntity as an abstract base class.

src/main/java/org/joshsim/engine/geometry/Grid.java
```diff
@@ -0,0 +1,45 @@
...
+
+
+  /**
+   * Constructor for the Grid class.
+   *
+   * @param patches List of patches in the grid.
+   * @param spacing Width of a cell in the grid.
+   */
+  public Grid(List<Patch> patches, EngineValue spacing) {
+    this.patches = patches;
```
sampottinger, line 24: For a later merge request, we might want to index into latitude and longitude with TreeMap.

sampottinger (approved):
Minor optional comments @GondekNP. Otherwise, looks good to me.

sampottinger:
@GondekNP - I went ahead and had spotless take care of most issues and manually resolved the rest. This is green. Merge when ready!

src/main/java/org/joshsim/engine/geometry/GeometryFactory.java
```diff
@@ -0,0 +1,87 @@
...
+  /**
+   * Creates a square geometry with the specified width and center.
+   *
+   * @param width The width of the square
+   * @param centerLatitude The center latitude
+   * @param centerLongitude The center longitude
+   * @return A Geometry object representing a square
+   */
+  public static Geometry createSquare(
+      BigDecimal width,
```
sampottinger, line 29: Sorry one more thought popped up. The only other thing I’m not sure about is if this could take Unit so we can specify size (not position) in meters or in degrees. It would be nice if this could handle that even if it is just a haversine. This could be here or it could be in another class that maybe adds that capability. Curious what you think @GondekNP.

sampottinger:
Sorry left one more comment about unit conversion.

src/main/java/org/joshsim/engine/geometry/GridBuilder.java
```diff
@@ -0,0 +1,161 @@
...
+      throw new IllegalArgumentException(
+        "Top-left longitude must be less than bottom-right longitude");
+    }
+  }
+
+  private List<Patch> createPatches() {
+    List<Patch> patches = new ArrayList<>();
+
+    // Convert cell width to degrees (assuming it's in the same units as lat/long)
+    double cellWidthDegrees = cellWidth.getAsDecimal().doubleValue();
```
sampottinger, line 116: Yeah, sorry I just followed that thread all the way through. I think we might need to deal with units here.

src/main/java/org/joshsim/engine/geometry/GeometryFactory.java
sampottinger, line 29: Sorry one more thought popped up. The only other thing I’m not sure about is if this could take Unit so we can specify size (not position) in meters or in degrees. It would be nice if this could handle that even if it is just a haversine. This could be here maybe as another overload or it could be in another class that maybe adds that capability. Curious what you think @GondekNP.

sampottinger (changes requested):
Sorry for the rigmarole. I might be missing something, but I wonder if we need to add in some additional logic for units, at least for degrees and meters.

src/main/java/org/joshsim/engine/geometry/GeometryFactory.java
GondekNP: Yes - I'm glad you brought this up, I was thinking about this as well. 

The way I see it, our users will not be expressing anything themselves in geographical / degree units other than perhaps the description of the grid. Once we get into describing the biology, we will be in some projected unit (seed dispersed 30m for eg). So while it would be nice to have this be unopinionated, I was wondering if the complexity is worth it in the end.

That said it doesn't change the fact that we should probably have `Units` here. Secondarily I was wondering if it would be sensible to return `EngineValue` from these methods (for eg `centerDistanceTo`) rather than `BigDecimal`, but was runnning into the issue of not having a `caster` to pass to the `EngineValue` unless we instantiate one with the `GeometryFactory` and pass it to the constructor of `Geometry`. Separate issue, but would allow us to use the casting / unit / dimensional analysis logic more directly?

src/main/java/org/joshsim/engine/geometry/GridBuilder.java
GondekNP: See below :)

src/main/java/org/joshsim/engine/geometry/GeometryFactory.java
sampottinger: These are very good points. I think maybe what the language side could do is decorate this so that everything gets standardized (the interpreter keeps track of what unit the grid is defined in and ensures that its decorated GeomteryFactory always converts to that). In that case, I think it might make sense to have the geometry stuff not interact with EngineValue and the JavaDoc would just refer to the grid units as opposed to degrees. What do you think?

src/main/java/org/joshsim/engine/geometry/GridBuilder.java
sampottinger: See below - would decorate on the language side. I think the only thing we need to do is change these to BigDecimal as opposed to EngineValue and ensure that the JavaDoc is generalized to units instead of (Earth) degrees.

src/main/java/org/joshsim/engine/geometry/GeometryFactory.java
GondekNP: That sounds good to me. So if I'm understanding this properly, will rely on the underlying Spatial4j logic to calculate units according to how it was instantiated (which as far as I know can be ascertained by the `SpatialContext` which should then be included as an overloaded constructor) - so then this just needs to be adequately generic in terms of var names and logic to return a BigDecimal either way.

GondekNP: And agreed to keep EngineValue stuff seperate!

sampottinger: Cool!

sampottinger (approved):
From my side, looks good to go ahead and merge when ready after the update to remove engine value!

src/main/java/org/joshsim/engine/geometry/GeometryFactory.java
```diff
@@ -0,0 +1,87 @@
...
+      BigDecimal centerLatitude,
+      BigDecimal centerLongitude
+  ) {
```
GondekNP, line 32: As per our discussion, to keep these methods static, we now require a spatial4j `SpatialContext` to be passed to the class method. However, we provide the default as well - see `GridBuilder` overloaded constructors.

sampottinger:
Yayy! Do you need any help getting over the finish line @GondekNP ?

sampottinger:
I’ll update #39.

PR #32, opened 2025-03-22
Add spotless to dev.

sampottinger:
Add spotless back into dev but using cleanthat instead of google-java-formatter to avoid overzealous edits to code that conforms to the google style guide.


sampottinger:
Superceeded by #33.

PR #33, opened 2025-03-22
Add spotless at simulation.

sampottinger:
So google-java-format is a bit overzealous on rewriting code that is compliant with the Google Style Guide. I think this is rewriting code though to a form that is less readable in certain cases, something it seems others have complained about in their issues. I’d prefer the balance between readability and consistency that the Google Style Guide promotes over that which is enforced by google-java-format. Therefore, I am proposing use of spotless but with cleanthat instead. This is much more conservative, compliant with the Google Style Guide, and we can add additional formatting rules on top (google-java-format says it is a feature not a bug that rules cannot be disabled for philosophical reasons).


sampottinger:
Superseded by #34 .

PR #38, opened 2025-03-24
Implementation of Converter.

sampottinger:
Implementation of Converter which creates transitive conversions.


sampottinger:
Adding tests now but opening up for review @GondekNP

sampottinger:
I think this is ready to go!

sampottinger:
Sorry about the noise there @GondekNP - just got this rebased against `dev`.

src/main/java/org/joshsim/engine/value/TransitiveConversion.java
```diff
@@ -0,0 +1,70 @@
...
+
+import org.joshsim.engine.func.CompiledCallable;
+import org.joshsim.engine.func.Scope;
+import org.joshsim.engine.func.SingleValueScope;
+
+
+/**
+ * A conversion rule between two unit types using two inner conversions.
+ */
+public class TransitiveConversion implements Conversion {
```
GondekNP, line 18: Since this converts only two steps... do we envision having some `base` unit that we expect will be a middle ground no matter what? For eg, you could imagine:
- mm to cm, cm to m, m to km - which would require this to use an iteratable of `Conversion` rather than `first` and `second` (unless there is a clever strat here)
- or, we assume all spatial units need to know how to convert to and from `m`, which would be in the current example, mm to m and then m to km. 

I think the second seems like a reasonable approach since this would not need refactor and in general, it seems reasonable to use a base unit of `m` for projected coordinates which is how spatial4j will see distances between patches.

GondekNP (commented):
LGTM - curious for your thoughts on 3+ step conversion, but going to prioritize getting dev merged up and green :heavy_check_mark:

PR #39, opened 2025-03-24
Add query cache for EngineBridge.

sampottinger:
Blocked on #26 and #31. CC @GondekNP


src/main/java/org/joshsim/lang/bridge/ShadowingEntity.java
```diff
@@ -219,6 +230,15 @@ public Simulation getMeta() {
     return meta;
   }
 
+  /**
```
GondekNP, line 233: (Optional) Not sure I love `getRaw` unless it means something in another context... I might prefer `getInner` or `getInnerEntity` or something like that?

GondekNP (approved):

src/main/java/org/joshsim/lang/bridge/ShadowingEntity.java
sampottinger: I'll change to getInner!

PR #43, opened 2025-03-24
Add value resolver.

sampottinger:
Added ValueResolver which can follow Josh language variable paths to get EngineValue within ANTLR interpreter.

### DIFF
--- a/src/main/java/org/joshsim/lang/interpret/ValueResolver.java
+++ b/src/main/java/org/joshsim/lang/interpret/ValueResolver.java
@@ -88,7 +88,8 @@ public class ValueResolver {
 
       String attemptPath = attemptJoiner.toString();
 
-      if (target.has(attemptPath)) {
+      try {
+        EngineValue value = target.get(attemptPath);
         foundPath = attemptPath;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/joshsim/lang/interpret/ValueResolver.java
```diff
@@ -7,22 +7,108 @@
...
+   * then recursively resolve any remaining path segments in nested entity scopes.</p>
+   *
+   * @param target The scope to search for the value.
+   * @return Optional containing the resolved value if found, empty otherwise.
    */
-  Optional<EngineValue> get(Scope target);
+  public Optional<EngineValue> get(Scope target) {
+    Optional<ValueResolver> continuationResolverMaybe = getInnerResolver(target);
+    if (continuationResolverMaybe == null) {
+      return Optional.empty();
```
GondekNP, line 51: Curious if we ever want to be returning `Null`- I haven't had a clear idea of the difference truly. For eg, `Geometry.createSquare` returns `null` if the square is not reasonably square - but wondering if it should return optional, or be called `createSquareMaybe` ?

PR #49, opened 2025-03-28
Feat/minio

GondekNP:
Implementation of Minio client to save outputs (but for now, just the src `.josh`) into a remote cloud bucket.


src/main/java/org/joshsim/util/MinioHandler.java
```diff
@@ -0,0 +1,156 @@
...
+      throw new IllegalArgumentException("MinioOptions does not contain a valid Minio URL");
+    }
+
+    this.minioClient = options.getMinioClient();
+    this.bucketName = options.getBucketName();
+    this.basePath = options.getObjectPath();
+    this.output = output;
+
+    // Ensure the bucket exists
+    ensureBucketExists();
```
sampottinger, line 42: I wonder if this could be behind a flag on command line. Like --ensure-bucket. Just unsure if we should do this without the user saying yes.

sampottinger (approved):
This is great! Just left a minor optional comment.

PR #51, opened 2025-03-30
Implementation of SingleThreadEventHandlerMachine

sampottinger:
Includes / supersedes #56.


sampottinger:
Writing tests now but opening for review. @GondekNP. I also don't think I need the getMaxTimestep anymore. I'll confirm and remove from Replicate.

Issue #52, opened 2025-03-30
ANTLR source generation doesn't go into proper subdir according to package declaration


GondekNP:
Within `build.gradle`:
```
// ANTLR configuration
generateGrammarSource {
  // Set output directory for generated files
  outputDirectory = file("src-generated/main/java")

  // Set package for the generated visitor and other classes
  arguments += ["-visitor"]

  // To use direct left recursion
  arguments += ["-Dlanguage=Java"]
}

```

This generates source into `src-generated/main/java/org/joshsim/lang`, rather than `src-generated/main/java/org/joshsim/lang/antlr` which is where it (probably?) ought to be according to its declaration: `package org.joshsim.lang.antlr;`.

Doesn't cause issues with build evidently, but causes `VSCode` to not be able to see the imports properly. After moving the generated source into the `antlr` folder manually, tests pass as expected. Unfortunately, it seems as though changing `outputDirectory` is only the prefix (it is "src-generated/main/java" after all).

GondekNP:
Closed by #53

PR #53, opened 2025-03-30
Feat/impl replicate

GondekNP:


GondekNP:
Fixes #52 - I've isolated the test issue to the location of the generated source files from `ANTLR`. There was a mismatch between the package declaration and where the `g4` actually was in the tree (it was in `src/main/antlr/org/joshsim/lang` rather than `src/main/antlr/org/joshsim/lang/antlr`). I opted to just move the file down rather than change the import paths, but perhaps it is better practice to simply change the package declaration in the `g4` - either works!

sampottinger (approved):

PR #63, opened 2025-04-03
Add distribution scope

sampottinger:
Allow for retrieving attributes across distribution of entities. Note that this also changes RealizedDistribution to throw an error if contents are requested beyond the size of that distribution without replacement. This makes it consistent with sample behavior.


GondekNP:
LGTM - sounds like the right time to throw anyways, so better in general

GondekNP (approved):
LGTM

PR #64, opened 2025-04-03
Add transitive freeze to Entity and EngineValue

sampottinger:


src/main/java/org/joshsim/engine/entity/base/MutableEntity.java
```diff
@@ -102,7 +102,20 @@ public void unlock() {
...
+    Map<String, EngineValue> frozenAttributes = new HashMap<>();
+    
+    for (String attributeName : attributes.keySet()) {
+      EngineValue unfrozenValue = attributes.get(attributeName);
+      EngineValue frozenValue = unfrozenValue.freeze();
+      frozenAttributes.put(attributeName, frozenValue);
+    }
+    
+    return new FrozenEntity(
+        getEntityType(),
```
GondekNP, line 114: Just had a thought - I can't really think of a reason right now, is there any reason to think a user would change `name`, `geometry` or `entitytype` during a sim? `EntityType` probably not, but trying to think if there would ever be a valid reason to update an entity's geometry or name (though name would be a pretty smelly thing to do). 

In general, wondering if we should just create freeze to be safe the other things that construct a FrozenEntity (though we might take a hit on lots of superfluous geometries).

GondekNP (approved):
LGTM - see comment on thinking through potential unexpected behavior....

src/main/java/org/joshsim/engine/entity/base/MutableEntity.java
sampottinger: This is a good point. I think, due to immutable Java types, I’m less worried about entity type and name. However, geometry could become a liability later if it becomes mutable. That said, I’m mixed on cloning geometry given the space overhead. I think, for now, I’ll document in Geometry that we expect it to be immutable.

sampottinger: Implemented. Thanks!

PR #66, opened 2025-04-03
Feat/cog external resource

GondekNP:
WithSubsetter

Sam and I worked out a strategy using a Decorator pattern (https://refactoring.guru/design-patterns/decorator) to isolate caching logic from the source of the external resource, abstracting all of this into a `Request` which does not care where the data came from and only wants a `RealizedDistribution`.


src/main/java/org/joshsim/engine/external/core/ExternalPathCacheLayer.java
```diff
@@ -0,0 +1,62 @@
...
+   */
+  public ExternalPathCacheLayer(ExternalLayer decoratedLayer) {
+    super(decoratedLayer);
+  }
+
+  @Override
+  public RealizedDistribution fulfill(Request request) {
+
+    // Check if we have this resource in cache
+    if (cache.containsKey(request)) {
```
sampottinger, line 34: I like this though I am not sure how large I want the cache to be. We might consider changing this to https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/map/LRUMap.html

src/main/java/org/joshsim/engine/external/core/PrimingGeometryLayer.java
```diff
@@ -0,0 +1,75 @@
...
+  public RealizedDistribution fulfill(Request request) {
+    // Get the geometry from the request
+    // TODO: Revisit proper times to throw here
+    Geometry requestGeometry = request.getGeometry().orElseThrow();
+
+    // Update our priming geometry to include this request
+    extendPrimingGeometry(requestGeometry);
+
+    // Pass the request to the decorated layer
+    return super.fulfill(request);
```
sampottinger, line 73: This looks good though I may be missing where fulfill is called with the priming geometry? I think it would be on line 72?

sampottinger (changes requested):
Looks great! Just two quick questions to check on that I think would be easy to address.

src/main/java/org/joshsim/engine/external/core/PrimingGeometryLayer.java
GondekNP: Ah yes my bad, this should have been a TODO as well and I'm glad you brought it up. 

Currently this is extending the priming geometry within the priming layer, but I wasn't exactly sure how we wanted to actually pass this off to the underlying layers. I'm expecting that you are going to make repeated calls with different small geometries that are part of the larger one, and then this layer will update the request with the rest of the geometries that are have been queried for this geometry before. 

However, there are maybe two issues:
1) If we have cogs per timestep, and we iterate through agents, it seems likely that we will be better off priming with the whole geometry (this is a solveable problem because it seems like the user or perhaps our static analysis(?) can determine if we should just prime with the whole geometry, but wondering if it is worth thinking about now or later.

2. To keep the logic sound, it seems like I would want to update the request with the primedGeometry for the purpose of getting the right cache, but I don't want to augment the request I assume. If that was preferable, we could do something like `request.setPrimedGeometry(), and then later layers can just use or ignore that as they see fit in various implementations. This would require an update to hashing logic (tbd) but would preserve the logic of just passing request to super. 

What do you think?

GondekNP: On 1 - assuming we do the convex hull and we activate agents serially per cell (?), perhaps this converges to the whole thing pretty quickly.

sampottinger: **For point 1:** Long term, I am tempted to write one more decorator that just takes in a geometry and uses that for priming no matter what. So maybe a ExtendingPrimingGeometryLayer and a StaticPrimingGeometryLayer. To your point, we could also give the user control over this behavior in the simulation given if they think it will be sparse or if it just makes sense to get the whole grid. I wonder if Joshua Tree may be a good use case for ExtendingPrimingGeometryLayer but I could imagine denser populations preferring StaticPrimingGeometryLayer.

**For point 2:** We could. I may be missing something obvious but I am not sure if I will know what the priming geometry should be from the interpreter side though. I would need to create the request at the agent-level.

I think, given the current hand we are playing, I would be tempted to leave this and potentially re-evaluate later so that the final integration branch can start. What do you think of renaming the current PrimingGeometryLayer to ExtendingPrimingGeometryLayer and ignoring these problems before maybe adding a StaticPrimingGeometryLayer that I just populate with the full grid? I can also add do that rename / add that if you would like outside this PR. Either way, I can plan to have a property I check to determine which one to use.

GondekNP: I agree - I think the move for now is to implement `StaticPrimingGometryLayer` and set that as default and kick the can down the road, since this also avoids the current issue of the convex hull / intersect spatial4j stuff not being sorted. In this case I would expect we get the PrimingGeometry from the `Simulation`?

src/main/java/org/joshsim/engine/geometry/Geometry.java
```diff
@@ -8,23 +8,33 @@
...
+  protected CoordinateReferenceSystem crs;
 
   /**
    * Constructs a Geometry with a provided spatial4j shape.
    */
-  public Geometry(Shape shape) {
+  public Geometry(Shape shape, CoordinateReferenceSystem crs) {
     this.shape = shape;
+    this.crs = crs;
+  }
```
GondekNP, line 31: Not sure if this is a can of worms... after writing a couple of tests, I was suprised that simply instantiating this with `crs = null` didn't cause a `NullPointerException`. Leaving this here because, we don't _want_ to allow a geometry without a CRS in my opinion, such that when we query an external layer we _know_ categorically we are doing the right transformation.

But, this begs the question - I have been using `Optional<Object>` pretty regularly, with the assumption that instantiating with `null` would throw. So - should we be propogating this `requireNonNull` elsewhere?

src/test/java/org/joshsim/engine/external/core/ExternalLayerFactoryTest.java
```diff
@@ -0,0 +1,386 @@
...
+  private EngineGeometry testArea2;
+  private EngineGeometry testAreaSmall;
+
+  @BeforeAll
+  static void setHeadlessMode() {
+    // Enable headless mode to avoid X11 dependencies
+    System.setProperty("java.awt.headless", "true");
+
+    // Disable native acceleration which can cause issues
+    System.setProperty("javax.media.jai.disableMediaLib", "true");
```
GondekNP, line 66: Note - I was seeing a lot of indications that JAI and 'native acceleration' caused issues running headless. It seems like the test failures we get (failure to import certain JAI libraries) is caused by not having this particular property. I suspect we might have to set these within the main application logic as well, but flagging this in case we want to sort out a way to do this more globably via an ENV var or similar.

build.gradle
```diff
@@ -33,11 +39,27 @@ dependencies {
...
+  // Geotools for Raster transformations and COG support
+  implementation("org.geotools:gt-main:32.2")
+  implementation("org.geotools:gt-geotiff:32.2")
+  implementation("org.geotools:gt-referencing:32.2")
+  implementation("org.geotools:gt-epsg-hsql:32.2")
 
-  // Apache SIS for coordinate system transformations and COG support
-  implementation("org.apache.sis.core:sis-referencing:1.4")
+  // JAI dependencies (required for Geotools)
+  implementation("javax.media:jai_core:1.1.3")
```
GondekNP, line 60: These dependencies also seemed necessary for avoiding these JAI-related issues. It's not exactly clear though to what degree JAI is necessary for headless settings

src/main/java/org/joshsim/engine/external/core/ExternalPathCacheLayer.java
```diff
@@ -0,0 +1,113 @@
...
+  public ExternalPathCacheLayer(ExternalLayer decoratedLayer) {
+    super(decoratedLayer);
+  }
+
+  @Override
+  public RealizedDistribution fulfill(Request request) {
+
+    if (request.getPrimingGeometry().isEmpty()) {
+      // If the request has no primingGeometry, we can just use the decorated layer
+      return super.fulfill(request);
```
GondekNP, line 43: This is a little bit iffy I think  - we essentially early return if we have any `primingGeometry` at all, which will always always be the case when we use the fully wrapped decorator. I know the better way to do this would be to rely on the base of the chain to be the request and then to have this level cache the request itself, but that was made a little awkward by the fact that we are not caching the request itself, we are caching the (transformed) `GridCoverage`. 

I suspect that we will want to add another level of caching here to check if, for a given `GridCoverage2D`, we've seen a given `Request` before - early indications are that interacting with the GridCoverage2D is not super fast (though the IO read / transform is really slow so that was a good thing to do upfront)

src/test/java/org/joshsim/engine/external/core/ExternalLayerFactoryTest.java
GondekNP: @sampottinger

src/main/java/org/joshsim/engine/external/core/ExternalPathCacheLayer.java
GondekNP: @sampottinger

src/main/java/org/joshsim/lang/bridge/GridFromSimFactory.java
```diff
@@ -74,14 +77,30 @@ public Grid build(MutableEntity simulation) {
...
+    CoordinateReferenceSystem targetCrsRef;
+    try {
+      // Check if the target CRS is valid
+      targetCrsRef = CRS.decode(targetCrs, true);
+    } catch (FactoryException e) {
+      throw new RuntimeException("Invalid target CRS: " + targetCrs, e);
+    }
+
     EngineValue sizeValueRaw = sizeMaybe.orElse(valueFactory.build(1, Units.COUNT));
     EngineValue sizeValue = convertToExpectedUnits(sizeValueRaw, Units.METERS);
```
GondekNP, line 97: @sampottinger Google style prefers we use `final` when the distance between some declaration and its usage is exceeded...

GondekNP:
@sampottinger - this is my `ExternalResource` stuff merged with all of the integration sprint changes to `dev`. I am planning to work on integrating minio and k8s stuff before circling back to netcdf, but it looks like it should be one more method to `ExternalLayerFacotory` with a different base layer.

In the current implementation, the cache layer is directly calling `CogReader` so they are not perfectly encapsulated from one another - I think the only way around it to cache the request-to-realized-distribution response (which I not what we want I don't think), or to have a chain for getting `GridCoverage2D` into memory and a chain for extracting from that to get `RealizedDistribution`. 

Also notable - no more `Spatial4j` or `Apache SIS` - all `geotools`, which made our lives easier in a few respects.

src/main/java/org/joshsim/engine/external/core/ExternalPathCacheLayer.java
sampottinger: Got it that’s a good flag. :( Let’s come back to it in a later merge request.

src/main/java/org/joshsim/engine/external/cog/CogReader.java
```diff
@@ -0,0 +1,144 @@
...
+
+  /**
+   * Read a coverage from a COG file for the specified geometry.
+   *
+   * @param path Path to the COG file
+   * @param geometry EngineGeometry defining the area of interest
+   * @return GridCoverage2D containing the data within the geometry's bounds
+   * @throws IOException if there is an error reading the file
+   */
+  public static GridCoverage2D getCoverageFromDisk(
```
sampottinger, line 35: Is it from disk? I would have expected the resources to be remote?

src/main/java/org/joshsim/engine/external/cog/CogReader.java
```diff
@@ -0,0 +1,144 @@
...
+      String path,
+      EngineGeometry geometry
+  ) throws IOException {
+    File file = new File(path);
```
sampottinger, line 39: If you want an easy way to support both, URL + BufferedReader + InputStreamReader might work. BufferedReader and InputStreamReader work the same way regardless of if the file is remote or local.

src/main/java/org/joshsim/engine/external/core/ExternalPathCacheLayer.java
```diff
@@ -0,0 +1,113 @@
...
+  /**
+   * Loads a GridCoverage from disk and caches it.
+   *
+   * @param path the path to the coverage file
+   * @param primingGeometry the geometry used to prime the cache
+   */
+  private void loadCoverageIntoCache(String path, EngineGeometry primingGeometry) {
+    GridCoverage2D newCoverage;
+    try {
+      newCoverage = CogReader.getCoverageFromDisk(path, primingGeometry);
```
sampottinger, line 82: What about network requests? Even if we don’t support them now, it might be good to name this something generic.

src/main/java/org/joshsim/lang/bridge/GridFromSimFactory.java
sampottinger: Sounds good

src/test/java/org/joshsim/engine/external/core/ExternalLayerFactoryTest.java
sampottinger: Sounds good!

sampottinger (approved):
Approved but curious about remote resources? We can do that in a follow up merge request but not sure if we might want to generalize some method names. Switching to a local file for a remote file is easy in Java. They can have the same interface.

src/main/java/org/joshsim/engine/external/cog/CogReader.java
sampottinger: Actually looks like Geotools takes either a URL or a File - https://github.com/geotools/geotools/blob/main/modules/plugin/geotiff/src/main/java/org/geotools/gce/geotiff/GeoTiffReader.java#L219. You might consider just using URL for local files too: `File.toURI().toURL()`

PR #67, opened 2025-04-03
Add synthetic scope for language keywords in interpreter side.

sampottinger:


src/main/java/org/joshsim/engine/value/type/LanguageType.java
```diff
@@ -19,26 +19,57 @@ public class LanguageType {
...
+  /**
+   * Creates a new LanguageType with distribution types and a root type without inner attributes.
    *
    * @param distributionTypes Collection of distribution type identifiers.
    * @param rootType The base type (e.g., "decimal", "string", etc.).
    */
   public LanguageType(Collection<String> distributionTypes, String rootType) {
     this.distributionTypes = distributionTypes;
     this.rootType = rootType;
+    this.containsAttributes = false;
```
GondekNP, line 57: So things like `IntScalar`, `RealizedDistribution` don't have attributes, things like `Patch` and `Agent` do?

GondekNP (commented):
Looks good thanks much.

PR #70, opened 2025-04-04
Integration sprint

sampottinger:


src/main/java/org/joshsim/engine/geometry/GridBuilderExtents.java
```diff
@@ -0,0 +1,104 @@
...
+ */
+
+package org.joshsim.engine.geometry;
+
+import java.math.BigDecimal;
+
+/**
+ * Structure to store and validate corner coordinates for grid construction.
+ */
+public class GridBuilderExtents {
```
GondekNP, line 15: Something for later, but having tinkered a bit more with Apache SIS - this probably should just be a thin wrapper around something like `Envelope` to make our lives easier

src/main/java/org/joshsim/engine/geometry/GridBuilderExtentsBuilder.java
```diff
@@ -0,0 +1,78 @@
...
+
+package org.joshsim.engine.geometry;
+
+import java.math.BigDecimal;
+
+
+/**
+ * Builder class for creating {@link GridBuilderExtents} instances.
+ */
+public class GridBuilderExtentsBuilder {
```
GondekNP, line 15: Also definitely something for later, but I wonder if it would make sense (from an organizational standpoint) to re-org our directories somewhat to group our `Builder`s since those are basically what enable the Josh `compiling` to work, but also would provide a direct interface to procedurally construct a simulation using the Engine directly. (you know how much I love re-orging folders :stuck_out_tongue_winking_eye: )

src/main/java/org/joshsim/engine/simulation/ReplicateBuilder.java
GondekNP, line 1: Not needed for the compiling process? I was figuring for direct engine use, it would be nice to be able to construct valid sims primarily using builders and constructors - wonder what you think about that

src/main/java/org/joshsim/lang/bridge/GridFromSimFactory.java
```diff
@@ -0,0 +1,188 @@
...
+import org.joshsim.engine.value.engine.EngineValueFactory;
+import org.joshsim.engine.value.type.EngineValue;
+import org.opengis.referencing.operation.TransformException;
+import org.opengis.util.FactoryException;
+
+
+/**
+ * Factory building a Grid from a simulation and bridge.
+ */
+public class GridFromSimFactory {
```
GondekNP, line 26: I wonder if we can integrate this better with `GridBuilder` or if it necessary to have this as a special factory meant just for the lang side

GondekNP:
LGTM Sam!!! This is a very exciting milestone. I made a couple of comments (I think most importantly, to what degree should we / how can we support building simulations directly from the Engine side) but I think we should touch base and just open a couple of issues for what things we want to improve (across the whole v0 at this point) rather than halt progress on learning what we will learn from getting things going.

src/main/java/org/joshsim/engine/geometry/GridBuilderExtents.java
GondekNP: https://sis.apache.org/apidocs/org.apache.sis.referencing/org/apache/sis/geometry/GeneralEnvelope.html

PR #85, opened 2025-04-13
Add CSV export for patches

sampottinger:
Adding CSV export just for patches for now to unblock folks. Will add for entity and simulation soon. CC @mzomer and @lucialayr. Note that this will write a CSV to `tmp` in integration test (`simple.josh`).


GondekNP (commented):
LGTM!

PR #100, opened 2025-04-23
Formalize a platform-specific input/output layer.

sampottinger:
Adding in InputOutputLayer to support different availability of import / export strategies depending on where the simulation is running (specifically JVM vs WASM).


GondekNP (commented):
LGTM!

PR #101, opened 2025-04-24
Feat/geo update patchset

GondekNP:
1. **Two-tier CRS approach**:
   - **Internal**: Grid-based coordinates (using indexes and a definition of the CRS to be instantiated on the geo side)
   - **External**: Earth-based coordinates (for visualization and API integration with spatial libraries `SIS`, `JTS` and `netcdf-java`)

2. **Transformation process**:
   - **Projected CRS**: Direct affine transformation (grid index  -> meters)
   - **Geographic CRS**: Two-step process:
     - Project to Transverse Mercator (centered at top-left)
     - Then apply affine transformation (grid index -> meters)


src/test/java/org/joshsim/lang/bridge/GridFromSimFactoryTest.java
```diff
@@ -73,32 +72,66 @@ void buildWithDefaultValues() {
...
+    when(mockSimulation.getAttributeValue("grid.low")).thenReturn(Optional.of(gridStartStr));
+    when(mockSimulation.getAttributeValue("grid.high")).thenReturn(Optional.of(gridEndStr));
+    when(mockSimulation.getAttributeValue("grid.size")).thenReturn(Optional.of(sizeVal));
+    when(mockSimulation.getAttributeValue("grid.patch")).thenReturn(Optional.empty());
+
+    PatchSet result = factory.build(mockSimulation);
+
+    assertNotNull(result, "PatchSet should be created with custom Transverse Mercator projection");
+    assertFalse(result.getPatches().isEmpty(), "PatchSet should contain patches");
+  }
```
GondekNP, line 136: These two tests currently throw because we don't support an EarthPatchBuilder - this is because I think it might be sensible to just generate them on the grid side by default and convert when necessary, rather than maintaining two seperate paths for generating grid cells. 

It's possible this is not very performant, since we'd be converting shape by shape, but I'm guessing we could stream it OR perhaps define some different route between the two.

src/test/java/org/joshsim/lang/bridge/GridFromSimFactoryTest.java
```diff
@@ -71,34 +62,71 @@ void buildWithDefaultValues() {
...
-    when(mockSimulation.getAttributeValue("grid.low")).thenReturn(Optional.of(gridStartStr));
-    when(mockSimulation.getAttributeValue("grid.high")).thenReturn(Optional.of(gridEndStr));
-    when(mockSimulation.getAttributeValue("grid.size")).thenReturn(Optional.of(sizeVal));
-
-    PatchSet result = factory.build(mockSimulation);
-
-    assertNotNull(result, "PatchSet should be created with custom values");
-    assertFalse(result.getPatches().isEmpty(), "PatchSet should contain patches");
-  }
+  // TODO: Determine whether we want to generate earth patches from earth side
```
GondekNP, line 65: This is a stub for a EarthPatchBuilder test, should we actually want to implement it

src/main/java/org/joshsim/geo/geometry/EarthPatchBuilder.java
```diff
@@ -1,310 +1,199 @@
...
-      throw new IllegalArgumentException(
-          "After transformation, top-left Y-coord must be greater than bottom-right Y-coord");
-    }
-
-    if (topLeftTransformed.getOrdinate(0) >= bottomRightTransformed.getOrdinate(0)) {
-      throw new IllegalArgumentException(
-          "After transformation, top-left X-coord must be less than bottom-right X-coord");
-    }
-  }
+public abstract class EarthPatchBuilder implements PatchBuilder {
```
GondekNP, line 11: As per the plan, `EarthPatchBuilder` is unimplemented for now, pending choices in the pre-processing pipeline. I don't think it would be hard to implement using `EarthGeometryFactory` and a similar spec to the Grid side, but the bigger question is - why would we do this? I suppose we could build Patches with EarthGeometry as their parent geom, but it seems like we would prefer to run the leaner engine side _Without_ explicit geometry now that they are decoupled, and adding `EarthPatchBuilder` means we need to worry about conversion the other way (from Earth to Grid)

GondekNP:
@sampottinger - We will talk about this on a huddle shortly, but big picture:

`GridCrsDefiniton` holds the necessary bits to define the grid CRS on the `geo` side

`GridCrsManager` doesn't actually instantiate a 'real' CRS anymore - this was causing all kinds of problems with operations / transforms, and required a lot of uselss-to-us metadata (since it's not a real CRS anyways). So instead, this now defines the necessary transforms to projected or geographic CRS

`EarthTransformer` is the single source for transformations both from `gridToEarth` and `earthToEarth`, using `GridCrsManager` to help translate. I used a ConcurrentHashMap here, assuming we cache the expensive transformation caclulations / lookups as a static class (which might be a bit Singleton, curious what you think, but seems helpful to house all transformations in one place)

`JtsTransformUtility` is mostly wrappers for out of the box transforms using the underlying JTS library (as well as figures how to instantiate geoms in grid space as actual geometries to be able to transform

Open questions are:

1) Do we need a EarthPatchBuilder? See comment.

2) How much imprecision is too much? We already see decent imprecision transforming back and forth from grid space, UTM, WGS84... see test's `delta` for various comparisons. I don't have a great sense of what is 'normal', but we should probably do some testing to figure out when this might be a problem (and figure out how much of it is perhaps floating point precision issues versus projection issues?)

3) Strategy to isolate building of the full 'geo' version versus the 'lite' sim version

GondekNP:
@sampottinger - see more editorializing in the merged-to-this #101

PR #112, opened 2025-04-30
Feat/geo external

GondekNP:
@sampottinger - more cleanup to be done, but netcdf logic is working as expected! :tada: 

A couple notes:
- See over at https://github.com/SchmidtDSE/josh-external-sources (`explotory/process_caladapt.py`) for how I queried from caladapt `climakitae`, converted and aggregated calendar year from `time_delta` and saved as netcdf (also saved test point csv".
- There are quite a lot of checks baked into the netcdf reader, for existence of various metadata and whatnot - this is for sure going to need to be revisited when we graduate from caladapt (we should give some more thought to the CF standard and what metadata we should expect in general
- I'm about to head to dinner sooo the csv points aren't validated yet, other than a smell check at the boundaries - will be done tonight
-


GondekNP:
And - one more - style pain aplenty here, I will revisit tonight and make sure this passes style as well!

PR #129, opened 2025-05-03
Threadsafe / streamable netcdf mapping to PatchSet

GondekNP:


sampottinger (approved):

src/main/java/org/joshsim/geo/external/ExternalGeoMapper.java
```diff
@@ -148,53 +166,151 @@ public Map<String, Map<Integer, Map<GeoKey, EngineValue>>> mapDataToPatchValues(
...
+   * This creates and manages its own data reader.
+   *
+   * @param dataFilePath Path to the data file
+   * @param variableName The variable to extract
+   * @param timeStep The time step to process
+   * @param patchSet Set of patches to map data to
+   * @return Stream of patch key to value entries that must be closed when done
+   * @throws IOException If there's an error reading the data file
+   */
+  public Stream<Map.Entry<GeoKey, EngineValue>> streamVariableTimeStepToPatches(
```
GondekNP, line 254: @sampottinger - see `streamVariableTimeStepToPatches`. I kept backwards compatability here for now, but we could envision deprecating the other methods here (that are not lazy and evaluate the stream directly).

sampottinger: Perfect - ship it!

PR #189, opened 2025-05-18
Fix some issues with slow string formatting.

sampottinger:
Found some time getting spent in `String.format` and an uncessary factory init. Altogether, shaves about 5% off in simple scripts though, in testing, this is shaving off up to 10% running time in scripts with lots of agents.


GondekNP (commented):
LGTM!

PR #249, opened 2025-08-13
Add support for config.

### LINKED ISSUE #248: Variables from config
# Variables from config

Allow loading variables from a config file which can be provided similar to jshd files.

## Background

Sometimes it is helpful to run simulations on different sets of parameters. These may be sets of different assumptions or possibly a specific intervention. Additionally, these parameters may come from a UI such as someone exploring sensitivities using sliders. These experiments may be conducted by the original author of the Josh simulation. However, they may also come from those less familiar with the platform and language.

Note that there is an "external data" feature which operates through jshd files generated from geospatial inputs like netCDF or geotiff. However, this is focused on raster data inputs with spatial and temporal specificity. In contrast, these configuration parameters are expected to have the same value over all time steps and over all spatial extents. Even so, external data may serve as inspiration for future work on configuration.

## Objective

Presently, the only way to change configuration is to change Josh code. This requires users to engage directly with the Josh language or, to support UI-based configuration (like sliders), one needs to perform text preprocessing.

To better facilitate this kind of experimentation and configuration, we should use a config prefix like `config configname.testVar`. This should find a `configname.jshc` file in working directory or on the request in the case of server-based execution containing lines like `testVar = 5 m` with one variable definition per line. These can be explicitly defined through a config stanzas or implied like seen in `grass_shrub_fire.josh`.

With this capability, users less familiar with Josh can more easily modify these parameters by editing these jshc text files but user interfaces or automation can also more readily generate these configurations to be read by Josh. However, to further facilitate use by other pieces of software, we should also add an option to determine which configuration variables are available.

## Implementation

This effort will require a series of components to be added or modified within Josh.

### ConfigInputParser

Right now there are classes reserved for future use for this purpose in `org.joshsim.engine.config`. We should convert `Config` and `ConfigBuilder` to concrete classes following the given interfaces. We should then add a `ConfigInputParser` which uses a `ConfigBuilder` to make a `Config` through a single public static method called `parse` taking in a single string. This should split by line and by equals sign such that statements like `testVar = 5 m` become a variable in the config (ex: `testVar` with value `5 m`).

In this specification, there must be exactly one variable per line with `#` used to make comments (all text after the hash mark ignored until the next newline. Empty lines are ignored. The only optional whitespace is any number (zero or more) spaces or tabs are allowed before and after the equals sign but, otherwise, no whitespace is allowed before the equals sign and the only whitespace after the equals sign is optional whitespace space between the number and the units for that EngineValue. This language is small enough that an ANTLR grammar should not be required and, instead, we can use an automaton similar to `UploadParserAutomaton` in [cdibase](https://raw.githubusercontent.com/sampottinger/cdibase/f4516277879fd6a96ca0e583aa49e1f75fb3eb16/prog_code/util/legacy_csv_import_util.py). Please credit cdibase in the README under open source used.

All of these classes will require new unit tests. For `ConfigInputParser`, please use the following example jshc file to be added to the examples directory:

```
# Group 1
testVar1 = 5 meters
test_var_2 =  10m

# Group 2
testVar3  = 15  km
```

This should result in 3 numbers on variables `testVar1`, `test_var_2`, and `testVar3`.

### Language support

Note that there is a prior design left as a stub for future use. Config stanzas should be similar to external stanzas but only need `source.location`. That said, like for external resources, config stanzas will be saved for future use. Even so, similar to the external keyword appearing in `expression`, we need a similar `configValue` rule similar to `externalValue` though there is no need for a `configValueAtTime` similar to `externalValueAtTime`.

Support will manifest within  `JoshParserToMachineVisitor` and support should be provided through a new `JoshConfigVisitor` similar to `JoshExternalVisitor`. Values are the same regardless of timestep and can be read similar to `pushExternal` but with a new method called `pushConfig` in `EventHandlerMachine` and implementers. This will require `getConfig` similar to `getExternal` on `EngineBridge` and implementers.

Tests for this component can be similar to external but please ensure the following:

 - Validate that scripts including the config stanza are valid.
 - Validate that scripts using the config keyword in expressions are valid.

Before starting this work, please review `JoshLang.g4` and `build.gradle` to ensure you understand how the ANTLR compilation works.

### Engine support for jshc

`ExternalResourceGetter` facilitates support for loading external resources. We should provide something similar called `ConfigGetter` (also in `org.joshsim.lang.bridge`). For now, we will just have `JshcConfigGetter` similar to `JshdExternalGetter`. This can also use `InputGetterStrategy`. That said, as values are unchanging, we should investigate if `JshcConfigGetter` can cache `Config` objects in a thread safe way.

Note that this requires updating `QueryCacheEngineBridge` and `MinimalEngineBridge` to take a new getter.  This requires updating construction of these two objects as well. However, the `JshcConfigGetter` can use the same input output layer and value factory as the `JshdExternalGetter`.

Testing should be similar to the `Jshd` equivalents.

### Editor support for jshc

The editor has a data files dialog at `#external-data-dialog`. Let's rename this to `#data-dialog`. Then, let's mention in the introductory `<p>` that this can include config files. Let's also mention there that jshd and jshc files are recommended. Let's update the JSDoc for `ExternalDataSerializer` to indicate that both config and geospatial data are sent through that object.

### Config discoverability

To assist other automated systems, we should offer support for determining what config variables are defined within a script. This should be a new visitor similar to `JoshParserToMachineVisitor` but without delegation. It should return a `Set<String>` that simply reports on when variables are used with a config. For example, `config configname.testVar` should report as `configname.testVar`. We should have a test which uses an example from the examples directory that has config included.

### Config from command line

We should make this functionality accessible through a new command similar to `RunCommand`. This should print a list of variables found one per line. See `JoshSimCommander`. This should be an action called `discoverConfig`.

### Config from editor

We should make a JS-accessible endpoint in `JoshJsSimFacade` with a new `@JSExport`. This should return back a string that is a comma separated list of variables found.

### Config dialog

We should make a new button similar to `#open-data-dialog-button` called `#open-config-dialog-button`. This should open a dialog similar to `#external-data-dialog`. In its introductory `<p>` we should say that variables defined here are accessible from `config editor.testVariable`. This dialog should have a textarea that defaults to just `testVariable = 5 m`. This should be included in the files sent with the request to the backend. To accomplish this task, let's have:

 - The dialog textarea read and edit a file via `LocalFileLayer`. Specifically, let's have it always read and write from / to `editor.jshc`.
 - The dialog should have save and cancel buttons where save will write via `LocalFileLayer`.
 - Upon opening the dialog through `#open-config-dialog-button`, the `editor.jshc` file current contents should be read. 

Here, simply checking the git diff to validate edits is sufficient for testing / verification.

### Documentation update

Please also update llms.txt to indicate support for configuration files. Tutorials will be updated in a later PR. The README.md is also incorrect about how jshd files work (currently looks in working directory). Please update information about jshd files are provided and then indicate that jshc files can be provided as well (mention `config from`). Finally, we can mention that external and config stanzas are reserved for future use. We should say that only the working directory / included files from the editor or HTTP request are currently released.

## Limitations

We should not make the UI for sliders or other types of inputs in this task. This will be left for a future effort. Additionally, we should not create production automation to generate jshc files. However, we should add an example to run as part of `validateExamples` (see `.github/workflows/build.yaml` that tests running with a configuration file).

## Notes

Note that, at this time, only jshd files from working directory are allowed. The external stanza is reserved for future use. Similarly, config stanzas will be saved for future use.

### LINKED ISSUE #253: Default values for config
# Default values for config

Option to add defaults values read from jshc configuration files.

## Background

Josh (#248) offers support for reading values from configuration files. This allows for users to evaluate sensitivity to different assumptions or run experiments without needing to engage with Josh Lang code directly for a simulation. This also provides affordances to users, indicating what parts of a simulation may be readily modified.

That said, there are limitations in the current approach. First, these values are all currently required. In other words, the config file must be passed in order for the simulation to run. Second, in allowing users and automation to discover configurable variables, we cannot specify what typical values might look like. This may reduce usability for automated systems trying to show UIs.

## Objective

To improve affordances for modifying values and to make it easier to pick up and run a simulation, we should add support to optionally specify default values for things that would have been read from a configuration file. For example:

```
height.init = config example.testVar1 else 5 m
height.init = config example.testVar2
```

Let's also aim to make the defaults available when inspecting configurable variables with output like so:

```
testVar1(5m)
testVar2
```

We can assume that defaults are given in parentheses and those without defaults simply omit that part of their description.

## Implementation

This requires updating / implementing a series of related components.

### Update Josh visitor

We should update `JoshLang.g4` to add support for the default (`else`) value. This reuses the `ELSE_` token. Then, we need to update `JoshParserToMachineVisitor`. Note that `EventHandlerMachine` has `pushConfig` but implementers largely call to `EngineBridge`. However, there are exceptions raised if the config data (like `configData`) doesn't have the named file or variable name. Perhaps `getConfig` may change to return an `Optional<EngineValue>` which is empty if not found and `pushConfig` can take an `Optional<EngineValue>` offering the default. If the default is not given and no value is found, we can raise an exception as a value is not available.

### Update config parser

We presently have a custom automaton in `ConfigInputParser`. However, as this config feature is getting larger, let's migrate this to ANTLR for consistency.

 - The visitor should go to `org.joshsim.lang.interpret.visitor`.
 - This should make a new g4 file, update the build scripts including GitHub actions.
 - We should replace `ConfigInputParser` (let's rename to `ConfigInterpreter`) with a new implementation with logic similar to seen in `JoshInterpreter`. This should be in `org.joshsim.lang.interpret`.
 - Create a new visitor for this new grammar in `org.joshsim.lang.interpret.visitor`.

This should use a fragment approach similar to `JoshParserToMachineVisitor`. Specifically, we should investigate if we can have a fragment with a number `BigDecimal`, a units `String`, an `EngineValue`, and a `ConfigBuilder`. Then, we can have a static method to combine `ConfigBuilder` instances into a new `ConfigBuilder` to support the combination process in ANTLR. For reference, see also `JoshConfigDiscoveryVisitor`. To disambiguate, let's rename the existing `Fragment` to `JoshFragment` in `org.joshsim.lang.interpret.fragment.josh` and make a new `JshcFragment` in `org.joshsim.lang.interpret.fragment.jshc`.

### Update discoverability visitor

We presently have `JoshConfigDiscoveryVisitor` which simply returns strings describing the variables found. However, we now need to parse the default values as well so that they can be provided during discovery commands. Similar to the updates to config parser, we should make a new `ConfigDiscoverabilityFragment` in `org.joshsim.lang.interpret.fragment.discover`. This can have a `String` for the name of the variable, `BigDecimal` for the number, a units `String`, an `EngineValue`, and a `Set` of a new object called `DiscoveredConfigVar` to be built in `org.joshsim.engine.config`. It should work similarly to `JoshFragment` and its implementers. Let's also ensure that `JoshConfigDiscoveryVisitor` is in `org.joshsim.lang.interpret.visitor`. Finally, a `DiscoveredConfigVar` in `org.joshsim.engine.config` should have a string representation available through `toString` used for hashing (regular Java overrides hash code) and equals (regular Java overrides equals) where two `DiscoveredConfigVar` instances are equal if their representation (like `DiscoveredConfigVar: testVar1(5m)` or `DiscoveredConfigVar: testVar2`) are the same. That way we can use a `HashSet`.

### Update discovery commands

There are multiple places where we report discovered variables including:

 - Command line: `DiscoverConfigCommand`
 - JSExport: `discoverConfigVariables` in `JoshJsSimFacade`
 - Server: `JoshConfigDiscoveryHandler`

We should update all of these to use the format described above. However, in the process, we may be seeing duplicate code to create this serialization (string format with one variable per line and the parentheses). We should make a `ConfigDiscoverabilityOuputFormatter` in `org.joshsim.engine.config` to help facilitate. This should take in an `Iterable` over `DiscoveredConfigVar` and produce a `String` which can be returned. Let's add a method to `DiscoveredConfigVar` called `describe` which returns a string like `testVar1(5m)` or `testVar2` which is used by `toString` but without the `DiscoveredConfigVar:` prefix.

### Other clean up

We have moved a few files around. Let's also move `NoOpConfigGetter` to `org.joshsim.engine.config`. Let's also move `JshcConfigGetter` to `org.joshim.engine.config`.

## Notes

When moving files, test files will need to be updated accordingly. Also, note that the config support is not yet merged to `main` or `dev` but is on  `feat/external-config`. Where you would have diffed against `main` or `dev`, please diff against `feat/external-config` to see changes instead. The default value should be used if the file is not given or if the file of the given name does not have the variable of the given name.

sampottinger:
Adds support for the external configuration functionality. This closes #248 which requires the introduction of the configuration option without default values. However, this also closes #253 which adds support for defaults if the configuration file is not given or does not contain the desired variable.


src/main/java/org/joshsim/lang/interpret/fragment/ProgramBuilder.java
```diff
@@ -45,7 +45,7 @@ public ProgramBuilder() {
    * @param fragment The fragment to add to the program
    * @throws IllegalArgumentException if the fragment type is not supported at the top level
    */
-  public void add(Fragment fragment) {
+  public void add(org.joshsim.lang.interpret.fragment.josh.JoshFragment fragment) {
```
sampottinger, line 48: Let's use `JoshFragment` with an import instead of fully qualified name `org.joshsim.lang.interpret.fragment.josh.JoshFragment`.

src/main/java/org/joshsim/lang/interpret/fragment/discover/ConfigValueFragment.java
```diff
@@ -0,0 +1,130 @@
...
+  private final Optional<EngineValue> engineValue;
+
+  /**
+   * Creates a new fragment for a simple name without additional value information.
+   *
+   * @param name The name string
+   * @throws IllegalArgumentException if name is null or empty
+   */
+  public ConfigValueFragment(String name) {
+    if (name == null || name.trim().isEmpty()) {
```
sampottinger, line 34: I think it's ok to drop this if statement and exception. It is inconsistent with the rest of the codebase.

src/main/java/org/joshsim/lang/interpret/fragment/discover/ConfigValueFragment.java
```diff
@@ -0,0 +1,130 @@
...
+
+  /**
+   * Creates a new fragment for a value with number and units.
+   *
+   * @param number The numeric value
+   * @param units The units string
+   * @throws IllegalArgumentException if number is null
+   */
+  public ConfigValueFragment(BigDecimal number, String units) {
+    if (number == null) {
```
sampottinger, line 51: I think it's ok to drop this if statement and exception. It is inconsistent with the rest of the codebase.

src/main/java/org/joshsim/lang/interpret/fragment/discover/ConfigValueFragment.java
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  /**
+   * Creates a new fragment for an engine value.
+   *
+   * @param engineValue The engine value
+   * @throws IllegalArgumentException if engineValue is null
+   */
+  public ConfigValueFragment(EngineValue engineValue) {
+    if (engineValue == null) {
```
sampottinger, line 67: I think it's ok to drop this if statement and exception. It is inconsistent with the rest of the codebase.

src/main/java/org/joshsim/lang/interpret/fragment/discover/ConfigVarSetFragment.java
```diff
@@ -0,0 +1,45 @@
...
+  private final Set<DiscoveredConfigVar> discoveredVars;
+
+  /**
+   * Creates a new fragment containing the specified set of discovered config variables.
+   *
+   * @param discoveredVars The set of discovered config variables
+   * @throws IllegalArgumentException if discoveredVars is null
+   */
+  public ConfigVarSetFragment(Set<DiscoveredConfigVar> discoveredVars) {
+    if (discoveredVars == null) {
```
sampottinger, line 30: I think it's ok to drop this if statement and exception. It is inconsistent with the rest of the codebase.

src/main/java/org/joshsim/lang/interpret/visitor/JoshConfigDiscoveryVisitor.java
```diff
@@ -0,0 +1,136 @@
...
+   *
+   * @param defaultValueContext The parse tree context for the default value expression
+   * @return A string representation of the default value
+   */
+  private String extractDefaultValueString(ParseTree defaultValueContext) {
+    if (defaultValueContext == null) {
+      return "";
+    }
+
+    // Handle simple cases based on context type using new-style switch
```
sampottinger, line 76: I don't believe this is new style switch.

src/main/java/org/joshsim/lang/interpret/visitor/JoshConfigDiscoveryVisitor.java
```diff
@@ -0,0 +1,136 @@
...
+    switch (defaultValueContext.getClass().getSimpleName()) {
```
sampottinger, line 77: Is there a way we can have the defaultValueContext accept us with visitor instead of this switch case on class name? We should avoid inspection if possible. Let's review the other ANTLR visitors in the project.

src/main/java/org/joshsim/lang/interpret/visitor/JoshConfigDiscoveryVisitor.java
```diff
@@ -0,0 +1,136 @@
...
+  private String extractUnitsValueString(JoshLangParser.UnitsValueContext unitsValueContext) {
+    if (unitsValueContext == null) {
+      return "";
+    }
+
+    StringBuilder result = new StringBuilder();
+    result.append(unitsValueContext.number().getText());
+
+    // Check for units - could be identifier or PERCENT_
+    if (unitsValueContext.identifier() != null) {
```
sampottinger, line 111: Is there a way that unitsValueContext can accept this instead of doing this kind of inspection?

src/main/java/org/joshsim/lang/interpret/visitor/JoshConfigDiscoveryVisitor.java
```diff
@@ -0,0 +1,136 @@
...
+
+/**
+ * Visitor to discover configuration variables used in Josh scripts.
+ *
+ * <p>This visitor traverses the AST and collects all configuration variable
+ * references found in config expressions. For example, "config example.testVar"
+ * would add "example.testVar" to the discovered set, while "config example.testVar else 5 m"
+ * would add "example.testVar(5 m)" to show the default value.</p>
+ */
+public class JoshConfigDiscoveryVisitor extends JoshLangBaseVisitor<Set<DiscoveredConfigVar>> {
```
sampottinger, line 18: Across this whole class, there are places where we are doing inspection or pseudo-inspection. We should do a web search and look at the other visitors in this project. Ideally, the use of the visitor pattern could handle this complexity without inspecting class names or checking if attributes are present. However, we should evaluate if the alternative if viable.

src/main/java/org/joshsim/lang/interpret/visitor/JoshConfigParserVisitor.java
```diff
@@ -0,0 +1,135 @@
...
+   * @param ctx The ANTLR config context
+   * @return A JshcFragment containing the complete configuration
+   */
+  @Override
+  public JshcFragment visitConfig(JoshConfigParser.ConfigContext ctx) {
+    ConfigBuilder builder = new ConfigBuilder();
+
+    // Process each config line
+    for (JoshConfigParser.ConfigLineContext lineContext : ctx.configLine()) {
+      if (lineContext.assignment() != null) {
```
sampottinger, line 51: Humm is there a way for lineContext to use `accept` on `this` as a visitor to avoid this kind of inspection? Perhaps `JshcFragment` could have an `getIsComment` method that, if true, we can ignore? We should do a web search and look at other ANTLR visitors in the project to see if there are other options. If we go this route, perhaps we add a `JshcCommentFragment` subclass.

src/main/java/org/joshsim/lang/interpret/visitor/JoshConfigParserVisitor.java
```diff
@@ -0,0 +1,135 @@
...
+        // Process assignment and add to main builder
+        String variableName = lineContext.assignment().ID().getText();
```
sampottinger, line 53: Depending on what we do with inspection, perhaps this would be a good place to do an `accept` on `this`?

src/main/java/org/joshsim/lang/interpret/visitor/JoshConfigParserVisitor.java
```diff
@@ -0,0 +1,135 @@
...
+   *
+   * @param ctx The ANTLR value context
+   * @return A JshcFragment containing the parsed value
+   */
+  @Override
+  public JshcFragment visitValue(JoshConfigParser.ValueContext ctx) {
+    String numberText = ctx.NUMBER().getText();
+    String unitsText = "";
+
+    if (ctx.ID() != null) {
```
sampottinger, line 92: Don't love this inspection. See about visitor pattern with `accept` and `this` if possible.

src/main/java/org/joshsim/lang/interpret/visitor/JoshConfigParserVisitor.java
```diff
@@ -0,0 +1,135 @@
...
+import org.joshsim.lang.interpret.fragment.jshc.JshcFragment;
+import org.joshsim.lang.interpret.fragment.jshc.JshcValueFragment;
+
+/**
+ * ANTLR visitor for parsing .jshc configuration files.
+ *
+ * <p>This visitor processes ANTLR parse trees generated from .jshc files
+ * and produces JshcFragment instances containing the parsed configuration data.</p>
+ */
+public class JoshConfigParserVisitor extends JoshConfigBaseVisitor<JshcFragment> {
```
sampottinger, line 26: Across this whole class, there are places where we are doing inspection or pseudo-inspection. We should do a web search and look at the other visitors in this project. Ideally, the use of the visitor pattern could handle this complexity without inspecting class names or checking if attributes are present. However, we should evaluate if the alternative if viable.

PR #252, opened 2025-08-14
Finish external config first draft

sampottinger:
Integrations which complete #249.


README.md
```diff
@@ -24,13 +24,25 @@ The easiest way to get started locally is to simply [get yourself a copy of open
 $ java -jar joshsim.jar run simulation.josh
 ```
 
-While COGs, geotiffs, and netCDF files can be provided directly, the preferred approach is to provide a jshd file which preprocesses these geospatial inputs for speed:
+While COGs, geotiffs, and netCDF files can be provided directly, the preferred approach is to provide a jshd file which preprocesses these geospatial inputs for speed. These jshd files are loaded from the working directory or included with the request:
 
 ```
 $ java -jar joshsim.jar preprocess simulation.josh MySimulation data.nc variable units output.jshd
 $ java -jar joshsim.jar run simulation.josh --data output.jshd
```
sampottinger, line 31: The data flag is not yet available? If not, let's leave this out of the example but maybe we can add a prior line:

```
$ ls output.jshd
```

README.md
```diff
@@ -24,13 +24,25 @@ The easiest way to get started locally is to simply [get yourself a copy of open
...
 ```
 
+Configuration files (.jshc) can be provided alongside simulations to parameterize behavior without modifying Josh code:
+
+```
+$ java -jar joshsim.jar run simulation.josh --config parameters.jshc
```
sampottinger, line 37: The config flag is not yet available? If not, let's leave this out of the example but maybe we can add a prior line:

```
$ ls parameters.jshc
```

README.md
```diff
@@ -59,7 +75,7 @@ You can run the local UI through [joshsim](https://language.joshsim.org/download
 $ java -jar joshsim.jar server
 ```
 
-This will start a local web server which makes the UI available via your browser where you can work in private. The local server supports a sandbox mode that limits access to only the code and jshd files provided, with no network access otherwise.
+This will start a local web server which makes the UI available via your browser where you can work in private. The local server supports a sandbox mode that limits access to only the code, jshd, and jshc files provided, with no network access otherwise.
```
sampottinger, line 78: > The local server supports a sandbox mode that limits access to only the code, jshd, and jshc files provided, with no network access otherwise.

Let's do the following:

```
The local server supports a sandbox mode that limits access to only the code, jshd, and jshc files provided. However, the sandbox disallows network access otherwise.
```

sampottinger (commented):
Self-review

PR #254, opened 2025-08-14
Add defaults to external config

### LINKED ISSUE #253: Default values for config

sampottinger:
Closes #253. Adds support for providing default values if the config file or variable are not given. These defaults can also be read through configuration discovery commands.


src/main/java/org/joshsim/engine/config/DiscoveredConfigVar.java
```diff
@@ -0,0 +1,122 @@
...
+  @Override
+  public boolean equals(Object obj) {
+    if (this == obj) {
+      return true;
+    }
+    if (obj == null || getClass() != obj.getClass()) {
+      return false;
+    }
+    DiscoveredConfigVar other = (DiscoveredConfigVar) obj;
+    return Objects.equals(name, other.name) && Objects.equals(defaultValue, other.defaultValue);
```
sampottinger, line 110: Let's use the getters for name and default value. Let's assign them to `otherName` and `otherDefaultValue`. Let's then have two booleans: `sameName` and `sameDefaultValue`. Then, we can do `return sameName && sameDefaultValue;`.

src/main/java/org/joshsim/lang/bridge/EngineBridge.java
```diff
@@ -167,12 +167,22 @@ public interface EngineBridge {
...
+
   /**
    * Get a configuration value by name.
    *
    * @param name The name of the configuration value to retrieve.
    * @return The configuration value as an EngineValue.
+   * @deprecated Use getConfigOptional instead
    */
+  @Deprecated
   EngineValue getConfig(String name);
```
sampottinger, line 186: This is new enough (not even on `main` or `dev`). Let's remove instead of deprecating.

src/main/java/org/joshsim/lang/bridge/MinimalEngineBridge.java
```diff
@@ -196,27 +196,37 @@ public EngineValue getExternal(GeoKey key, String name, long step) {
...
+          configData.put(configName, configGetter.getConfig(configName));
+        } catch (Exception e) {
+          return Optional.empty();
+        }
+      }
+
+      Config config = configData.get(configName);
+      EngineValue value = config.getValue(variableName);
+      return Optional.ofNullable(value);
+    } catch (Exception e) {
```
sampottinger, line 220: Let's avoid these blanket try / catch with Exceptions. Why is this appearing and can we deal with it somehow else?

src/main/java/org/joshsim/lang/bridge/MinimalEngineBridge.java
```diff
@@ -196,27 +196,37 @@ public EngineValue getExternal(GeoKey key, String name, long step) {
...
+      if (parts.length != 2) {
+        return Optional.empty();
+      }
+      String configName = parts[0];
+      String variableName = parts[1];
+
+      if (!configData.containsKey(configName)) {
+        try {
+          configData.put(configName, configGetter.getConfig(configName));
+        } catch (Exception e) {
```
sampottinger, line 212: Let's avoid these blanket try / catch with Exceptions. Why is this appearing and can we deal with it somehow else?

src/main/java/org/joshsim/lang/interpret/ConfigInterpreter.java
```diff
@@ -0,0 +1,68 @@
...
+  public Config interpret(String configContent, EngineValueFactory valueFactory) {
+    try {
+      // Create lexer and parser
+      JoshConfigLexer lexer = new JoshConfigLexer(CharStreams.fromString(configContent));
+      CommonTokenStream tokens = new CommonTokenStream(lexer);
+      JoshConfigParser parser = new JoshConfigParser(tokens);
+
+      // Add error handling
+      parser.removeErrorListeners(); // Remove default console error listener
+      parser.addErrorListener(new org.antlr.v4.runtime.BaseErrorListener() {
```
sampottinger, line 44: Let's pull this out into a named inner class.

src/main/java/org/joshsim/lang/interpret/ConfigInterpreter.java
```diff
@@ -0,0 +1,68 @@
...
+      // Parse the config content
+      ParseTree tree = parser.config();
+
+      // Create visitor and visit the parse tree
+      JoshConfigParserVisitor visitor = new JoshConfigParserVisitor(valueFactory);
+      JshcFragment fragment = visitor.visit(tree);
+
+      // Extract Config from the resulting fragment
+      return fragment.getConfigBuilder().build();
+    } catch (Exception e) {
```
sampottinger, line 63: Let's avoid these blanket try / catch with Exceptions. Why is this appearing and can we deal with it somehow else?

src/main/java/org/joshsim/lang/interpret/fragment/discover/ConfigValueFragment.java
```diff
@@ -0,0 +1,140 @@
...
+   *
+   * @param name The name string
+   * @throws IllegalArgumentException if name is null or empty
+   */
+  public ConfigValueFragment(String name) {
+    if (name == null || name.trim().isEmpty()) {
+      throw new IllegalArgumentException("Name cannot be null or empty");
+    }
+    this.name = name.trim();
+    this.number = null;
```
sampottinger, line 37: Instead of null and null checks, let's use `Optional`.

src/main/java/org/joshsim/lang/interpret/fragment/discover/ConfigVarSetFragment.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+    this.discoveredVars = discoveredVars;
+  }
+
+  /**
+   * Gets the set of discovered config variables from this fragment.
+   *
+   * @return Set of discovered config variables
+   */
+  @Override
```
sampottinger, line 41: For `Override`-tagged methods, we do not need to repeat JavaDoc.

src/main/java/org/joshsim/lang/interpret/fragment/discover/ConfigVarSetFragment.java
```diff
@@ -0,0 +1,55 @@
...
+  public Set<DiscoveredConfigVar> getDiscoveredConfigVars() {
+    return discoveredVars;
+  }
+
+  /**
+   * Gets the type of this fragment.
+   *
+   * @return CONFIG_VAR_SET fragment type
+   */
+  @Override
```
sampottinger, line 51: For `Override`-tagged methods, we do not need to repeat JavaDoc.

src/main/java/org/joshsim/lang/interpret/fragment/jshc/JshcConfigBuilderFragment.java
```diff
@@ -0,0 +1,51 @@
...
+    this.configBuilder = configBuilder;
+  }
+
+  /**
+   * Gets the ConfigBuilder from this fragment.
+   *
+   * @return The ConfigBuilder instance
+   */
+  @Override
+  public ConfigBuilder getConfigBuilder() {
```
sampottinger, line 37: For `Override`-tagged methods, we do not need to repeat JavaDoc.

src/main/java/org/joshsim/lang/interpret/fragment/jshc/JshcConfigBuilderFragment.java
```diff
@@ -0,0 +1,51 @@
...
+    return configBuilder;
+  }
+
+  /**
+   * Gets the fragment type.
+   *
+   * @return The fragment type (PROGRAM)
+   */
+  @Override
+  public FragmentType getFragmentType() {
```
sampottinger, line 47: For `Override`-tagged methods, we do not need to repeat JavaDoc.

src/main/java/org/joshsim/lang/interpret/fragment/jshc/JshcFragment.java
```diff
@@ -0,0 +1,87 @@
...
+   * similar to the pattern used in JoshFragment. Note: Since ConfigBuilder doesn't
+   * expose its internal values, this method creates a new builder and assumes
+   * the caller will handle the merging through the visitor pattern.</p>
+   *
+   * @param first The first ConfigBuilder
+   * @param second The second ConfigBuilder
+   * @return A new ConfigBuilder for combining (implementation deferred to usage)
+   */
+  public static ConfigBuilder combineConfigBuilders(ConfigBuilder first, ConfigBuilder second) {
+    // Since ConfigBuilder doesn't expose its values, we return a new builder
```
sampottinger, line 82: Let's change this (see also visitor to see if updates are needed) such that `ConfigBuilder` has a method called `combine` which takes another `ConfigBuilder` and returns a new `ConfigBuilder` with combined values.

src/main/java/org/joshsim/lang/interpret/visitor/JoshConfigDiscoveryVisitor.java
```diff
@@ -22,27 +25,100 @@ public JoshConfigDiscoveryVisitor() {
...
+   * @param defaultValueContext The parse tree context for the default value expression
+   * @return A string representation of the default value
+   */
+  private String extractDefaultValueString(ParseTree defaultValueContext) {
+    if (defaultValueContext == null) {
+      return "";
+    }
+
+    // Handle simple cases based on context type
+    if (defaultValueContext instanceof JoshLangParser.SimpleExpressionContext) {
```
sampottinger, line 76: Don't love this instanceof. Is there another option in ANTLR? Also, can we switch to new style switch statement?

src/main/java/org/joshsim/lang/interpret/visitor/JoshConfigParserVisitor.java
```diff
@@ -0,0 +1,138 @@
...
+      // Create units object
+      Units units = unitsText.isEmpty() ? Units.EMPTY : Units.of(unitsText);
+
+      // Create EngineValue
+      EngineValue engineValue = valueFactory.parseNumber(numberText, units);
+
+      return new JshcValueFragment(number, unitsText, engineValue);
+    } catch (NumberFormatException e) {
+      throw new IllegalArgumentException("Invalid number format: " + numberText, e);
+    } catch (Exception e) {
```
sampottinger, line 109: Don't love the catch of all Exceptions. Is there another way to do this or is this needed? If so, why is it needed?

src/main/java/org/joshsim/lang/interpret/visitor/JoshConfigParserVisitor.java
```diff
@@ -0,0 +1,138 @@
...
+import org.joshsim.lang.interpret.fragment.jshc.JshcFragment;
+import org.joshsim.lang.interpret.fragment.jshc.JshcValueFragment;
+
+/**
+ * ANTLR visitor for parsing .jshc configuration files.
+ *
+ * <p>This visitor processes ANTLR parse trees generated from .jshc files
+ * and produces JshcFragment instances containing the parsed configuration data.</p>
+ */
+public class JoshConfigParserVisitor extends JoshConfigBaseVisitor<JshcFragment> {
```
sampottinger, line 26: Across this, let's use `ctx.*.accept` instead of calling the visit directly similar to `JoshMathematicsVisitor` and others.

PR #258, opened 2025-08-20
Remote run from jar

### LINKED ISSUE #257: Run remote from jar
# Allow remote execution from jar

Allow the jar file to allow remote execution from a local jar.

## Background

There are a few different ways to execute a Josh simulation on various back-ends and locations:

| Client           | Run WASM          | Run Local       | Run Remote          |
| ---------------- | ----------------- | --------------- | ------------------- |
| Editor (Browser) | WasmEngineBackend | _detectLocalApp | RemoteEngineBackend |
| Jar (CLI)        | Not Desirable     | RunCommand      | Not Supported       |

When running remote, this goes through `JoshSimLeaderHandler` which serializes to a "wire" format written by `MapToMemoryStringConverter` (see also `SandboxInputOutputLayer`) and read by `wire.js`. See also `org.joshsim.lang.io`.

Related to this, the path for serialization goes through `ExportSerializeStrategy` which converts `Entity` to a serializable value, in this case `Map<String, String>` per `MapExpertSerializeStrategy`.

## Objective

We should implement the ability to run a Josh simulation at a remotely running `JoshSimServer` from a local jar via the CLI. This involves clarifying which data files are to be transmitted in the request because sending everything in the working directory may be too much.

## Implementation

This feature requires implementing multiple components.

### Component 1: Expand memory string converter

We should rename  `MapToMemoryStringConverter` to `WireConverter` and `convert` to `serializeToString`. Then, let's make a `NamedMap` which is just an object with getters for final `String name` and `Map<String, String> target`. This can be the new input to `serializeToString`. We should remove (not deprecate) the old signature. Then, we should add a `deserializeFromString` method to parse to a `NamedMap`.

### Component 2: Local export

Having a `NamedMap`, we need to be able to export to `CsvExportFacade`, `GeotiffExportFacade`, and `NetcdfExportFacade`. We should implement this by adding another overload to `write` on `ExportFacade` that writes a `NamedMap` instead of an `Entity`. This intersects with some threading logic but `Task` can change to have `Optional<Entity>` and `Optional<NamedMap>` with two constructors, one setting the `entity` to empty and the other setting `namedMap` to empty. THe getters can return `Optional` as well. Adding this will allow wire deserializers to avoid going all the back to `Entity`.

### Component 3: Jar specify external files

There are two types of external files currently support: `jshd` and `jshc`. Currently, when running from the jar, we load via `JvmInputGetter` which primarily uses `loadFromWorkingDir`. This is fine for local execution as the files do not need to be sent over network. However, serializing everything can cause issues when using `SandboxInputGetter` as the request may be quite large. We should change `JvmInputGetter` to an abstract base class. Then, we can have subclasses `JvmWorkingDirInputGetter` use `loadFromWorkingDir` but `JvmMappedInputGetter` take in a list of file paths where a `Map<String, String>` maps from filename to path. In this second child class, we can then replace `loadFromWorkingDir` with `loadFromMappedFiles` though the existing `loadFromUri` may remain in place (inherited from the base class). We can achieve this by making `JvmInputGetter` a template method class with a abstract method `readImpliedPath` where that calls either `loadFromWorkingDir` or `loadFromMappedFiles`. For the mapping version, we should have this string built up from `--data` which can specify files to be in scope for a `RunCommand`. We can then have an `InputOutputLayer` parameter on `getJoshProgram` and have the `Command` classes instantiate a `JvmInputGetter` child. For all commands but `RunCommand` we can use `JvmWorkingDirInputGetter`. For `RunCommand`, we will use `JvmWorkingDirInputGetter` if no `--data` given. Otherwise, will use `JvmMappedInputGetter`. These classes will require unit tests. We should update the github actions to have an example with external data run both with and without `--data`.

### Component 4: RunRemoteCommand with arbitrary remote

Similar to `RunCommand` but backed by logic similar to `RemoteEngineBackend`, we should support execution of running with a remote machine which has a `JoshSimServer`. We should ask the user to specify the `--endpoint` and `--api-key`. This should use HTTP streaming. We can deserialize from the wire type (see `wire.js`) to a `NamedMap`. Then, we can use the overload for `write` to persist the results.

### Component 5: RunRemoteCommand with Josh Cloud

Similar to how we have the Josh Cloud option (`engine-josh-cloud`) , we should have `--endpoint` default to the Josh Cloud end point if not given. See `DEFAULT_ENDPOINT` in `run.js`.

sampottinger:
Allow using a local jar to submit jobs to a remote jar. Closes #257.


src/main/java/org/joshsim/util/ProgressCalculator.java
```diff
@@ -0,0 +1,242 @@
...
+   */
+  private String formatProgressMessage(double percentage, long currentStepInReplicate, 
+                                      int currentReplicate, boolean isReplicateComplete) {
+    if (totalReplicates == 1) {
+      // Single replicate: "Progress: 45.2% (step 123/500)"
+      return String.format("Progress: %.1f%% (step %d/%d)", 
+          percentage, currentStepInReplicate, totalStepsPerReplicate);
+    } else {
+      // Multiple replicates: "Progress: 45.2% (step 123/500, replicate 2/10)"
+      String base = String.format("Progress: %.1f%% (step %d/%d, replicate %d/%d)",
```
sampottinger, line 173: This is pretty long, let's do one parameter per line.

src/main/java/org/joshsim/util/ProgressCalculator.java
```diff
@@ -0,0 +1,242 @@
...
+   * @param currentStepInReplicate The current step within the replicate
+   * @param currentReplicate The current replicate number
+   * @param isReplicateComplete Whether this is a replicate completion message
+   * @return Formatted progress message string
+   */
+  private String formatProgressMessage(double percentage, long currentStepInReplicate, 
+                                      int currentReplicate, boolean isReplicateComplete) {
+    if (totalReplicates == 1) {
+      // Single replicate: "Progress: 45.2% (step 123/500)"
+      return String.format("Progress: %.1f%% (step %d/%d)",
```
sampottinger, line 169: For consistency with the call below, let's do one parameter per line.

src/main/java/org/joshsim/util/SimulationMetadataExtractor.java
```diff
@@ -0,0 +1,186 @@
...
+ */
+public class SimulationMetadataExtractor {
+
+  /**
+   * Container class for simulation metadata.
+   *
+   * <p>This class encapsulates key simulation parameters extracted from Josh scripts,
+   * providing easy access to step range information needed for progress calculations.</p>
+   */
+  public static class SimulationMetadata {
```
sampottinger, line 46: Pull out into own file.

src/main/java/org/joshsim/util/WireResponseParser.java
```diff
@@ -0,0 +1,235 @@
...
+    ERROR
+  }
+
+  /**
+   * Container class for parsed response data.
+   *
+   * <p>This class encapsulates the different types of information that can be
+   * contained in a streaming response line from a remote Josh server.</p>
+   */
+  public static class ParsedResponse {
```
sampottinger, line 53: Move into its own file also in `org.joshsim.wire`

src/main/java/org/joshsim/util/WireResponseParser.java
```diff
@@ -0,0 +1,235 @@
...
+     *
+     * @return The error message, or null if not an ERROR response
+     */
+    public String getErrorMessage() {
+      return errorMessage;
+    }
+
+    @Override
+    public String toString() {
+      return String.format("ParsedResponse{type=%s, replicate=%d, step=%d, data='%s', error='%s'}",
```
sampottinger, line 163: Kinda long, let's do one parameter per line.

src/main/java/org/joshsim/util/WireResponseParser.java
```diff
@@ -0,0 +1,235 @@
...
+   *   <li>[error] message - An error occurred with the given message</li>
+   *   <li>[N] data - Data point from replicate N</li>
+   *   <li>[N] - Empty data point from replicate N (ignored)</li>
+   * </ul>
+   *
+   * @param line The response line to parse
+   * @return A ParsedResponse object, or null if the line should be ignored
+   * @throws IllegalArgumentException if the line format is invalid
+   */
+  public static ParsedResponse parseEngineResponse(String line) {
```
sampottinger, line 185: Let's break this into a series of `WireResponseParserStrategy` implementers. Let's have one per matcher we have here. Let's have an `Optional<WireParseResult>` returned by each to avoid returning null. Let's have it be empty if it doesn't apply. Let's make a list of these and apply until one returns a non-empty response. We can have a flag for the line to be ignored. We can change the return value of `parseEngineResponse` to be `Optional<ParsedResponse>`

src/test/java/org/joshsim/command/RunRemoteCommandTest.java
```diff
@@ -0,0 +1,352 @@
...
+import java.util.Map;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+
+/**
+ * Test cases for RunRemoteCommand functionality.
+ */
+public class RunRemoteCommandTest {
```
sampottinger, line 29: This is great but let's be very careful that this is not actually generating network traffic.

src/test/java/org/joshsim/lang/io/JvmInputGetterTest.java
```diff
@@ -44,4 +46,16 @@ void testOpenFileFromWorkingDirectory() throws Exception {
...
+  @Test
+  void testExistsFileFromWorkingDirectory() {
+    // Act & Assert
+    assertTrue(inputGetter.exists(TEST_FILE_PATH));
+  }
+
+  @Test
+  void testUriHandling() {
+    // Act & Assert - URIs should be assumed to exist (they're handled by loadFromUri)
+    assertTrue(inputGetter.exists("http://example.com/test.txt"));
```
sampottinger, line 59: Looks good but let's check that this is not generating real network traffic.

src/test/java/org/joshsim/lang/io/JvmWorkingDirInputGetterTest.java
```diff
@@ -0,0 +1,76 @@
...
+    // Act & Assert
+    assertThrows(RuntimeException.class, () -> {
+      inputGetter.open("/tmp/non-existent-file.txt");
+    });
+  }
+
+  @Test
+  void testUriExists() {
+    // URIs should always return true for exists check
+    assertTrue(inputGetter.exists("http://example.com/test.txt"));
```
sampottinger, line 73: Let's please confirm that this is not generating real network traffic.

PR #260, opened 2025-08-21
Fix `readValuesAt` imprecision

GondekNP:
Unclear what the reason was, but the `readValuesAt` logic was previously rounding values at 6 digits (and also rounding up). Can't really think of a good reason why this existed in the first place - by NOT doing that, we get back to the precision inherent in writing to netcdf (IEEE 754?)!


src/main/java/org/joshsim/geo/external/readers/NetcdfExternalDataReader.java
```diff
@@ -184,6 +184,37 @@ public List<String> getVariableNames() throws IOException {
     return result;
   }
 
+  /**
+   * Retrieves the time dimension values from the NetCDF file.
+   *
+   * @return List of time dimension values, or empty list if not available
+   * @throws IOException If the NetCDF file is not open or an error occurs while reading
+   */
+  public List<Double> getTimeDimensionValues() throws IOException {
```
GondekNP, line 193: There was also an issue in that the centered year in the caladapt data was previously hardcoded at 2025, but in fact was 2024, leading to an off-by-one. Just an issue in the test due to my negligence but added this to be more robust down the line for arbitrary netcdf files

src/main/java/org/joshsim/geo/external/readers/NetcdfExternalDataReader.java
```diff
@@ -406,7 +437,7 @@ public Optional<EngineValue> readValueAt(
 
       // Create units and BigDecimal value
       Units units = Units.of(unit);
-      BigDecimal bigDecimalValue = new BigDecimal(value).setScale(6, RoundingMode.HALF_UP);
```
GondekNP, line 409: This is the offending line

PR #286, opened 2025-10-15
Minio streaming implementation

GondekNP:
Enabled sim results to stream directly to MinIO storage, as was our original goal with `minio://path/to/file` syntax in the josh language. 

Some key changes:
- Changed to `LinkedBlockingQueue` within consumer process to avoid heap exhasution due to slow writes - can be configured with `--export-queue-size` to tune, if we still exhaust heap
- Implemented `MinioOutputStrategy` using piped streams to get around a limitation in `minio-java` - see https://github.com/minio/minio-java/issues/956
- Basic exponential backoff for network failures, with eventually the simulation halting if minio can't be reached
- Implemented a Singleton `MinioClient` for (hopefully) efficient connection pooling - since we have a single consumer thread at the moment Singleton seems okay, but maybe update one day if we prefer otherwise
- Lastly, implemented an integration test which spins up a lil minio server in github actions, isolated from the rest of the build action

### DIFF
--- a/src/main/java/org/joshsim/command/RunCommand.java
+++ b/src/main/java/org/joshsim/command/RunCommand.java
@@ -366,12 +366,10 @@ public class RunCommand implements Callable<Integer> {
             parsedOutputSteps
         );
 
-        // Report replicate completion (except for the last simulation)
-        if (totalSimulationCount < jobs.size() * replicates) {
-          ProgressUpdate completion = progressCalculator.updateReplicateCompleted(
-              totalSimulationCount);
-          output.printInfo(completion.getMessage());
-        }
+        // Report replicate completion
+        ProgressUpdate completion = progressCalculator.updateReplicateCompleted(
+            totalSimulationCount);
+        output.printInfo(completion.getMessage());
       }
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/joshsim/lang/io/ReplicateOutputStreamGenerator.java
```diff
@@ -44,57 +44,61 @@ public class ReplicateOutputStreamGenerator implements
...
+   * Creates an OutputStream for the specified path using the strategy factory.
    *
    * @param path The file path to create stream for
-   * @return New FileOutputStream for the specified path
-   * @throws RuntimeException if the file cannot be created
+   * @return New OutputStream for the specified path
+   * @throws RuntimeException if the stream cannot be created
    */
-  private OutputStream createFileOutputStream(String path) {
+  private OutputStream createOutputStream(String path) {
```
GondekNP: Pass the ExportTarget - don't use smelly Function<> strategy factory, if we pass the ExportTarget we will know which output stream strategy to use and can type it properly

PR #318, opened 2025-11-25
Feat/graph viz tools

GondekNP:
This PR adds the ability to track and visualize dependency graphs from Josh simulations, by (1) capturing dependencies through the exisitng visitor pattern during interpretation, (2) extracting the graph structure from that tracked dependency information, and (3) visualized in an interactive graph structure (which for now is very dev-focused but could one day provide some tooling for more ecologically relevant insight). 

It can be run with the following:
```
  java -jar joshsim.jar exportDeps my_simulation.josh MySimulation -o
  deps.json
```

For eg, `test_states_basic.josh`:
image

Hover text provides additional context on which lines generated the dependency:
image


GondekNP:
Low priority for now and long out of date. I think eventually we might want some on the fly dependency graph auditing, but closing for now.

PR #326, opened 2025-12-01
[dev-restore] Debug output logic

GondekNP:
Mirror of #305 against clean `dev-restore` branch. `dev-restore` is meant to exclude a couple of bugs introduced to `dev` (#311 and likely #315), but include debug logging feature (#305), templating updates (#308), integration / conformance test suite (#310), random seed logic (#316), and documentation updates (#307).


GondekNP (commented):
Okay - after talking it over with @lucialayr - we are going to revert `GeoKey` back to its original state and remove all of the extraneous plumbing of threadsafe sequence ID. 

This was orignally set up under a false idea of overhead - bascially by iterating on GeoKey, we could have a unique combo of `GeoKey x sequenceId` to identify an agent, which would be consistent across simulations when seed is fixed. However, this held some appeal originally when I thought this would be a performance improvement over just having a completely seperate index for agents, but in fact we are essentially creating a new `GeoKey(sequenceId)` for each id, defeating the purpose and definitely worse off because all of the `GeoKey` bytes are duplicated. 

After talking with Lucia, we determined there isn't (probably) a strong case for being able to track a single agent across different replicates (as in, being able to isolate the `agent #1 in geokey(1,1)`), since within-replicate interactions are much more interesting and useful. Plus, depending on what else has changed, the 'same agent' might be initialized with different attributes, different external data, etc - is it even the same agent anymore?

Instead we will just use a lazy, global `AtomicLong`, which will be globally unique and may change from simulation to simulation because the exact ordering of thread execution is not consistent. But, it will dramatically reduce the amount of plumbing we need around the codebase whilst providing most of the utility that we need for debugging.

GondekNP:
Due to merge conflict hell (and due to some overengineering reflection) the debug feature from this branch has been deprecated for #334

Issue #322, opened 2025-11-30
Bug - invalid attribute access error when missing a state handler


GondekNP:
Related to #321 - if we start in a state that doesn't have a valid state handler (as in below, we have no seedling handler), we get an `Unable to get value for ValueResolver(Trees.age)`. While I wouldn't expect the below to be interesting, I would expect it to run and contain only seedlings, however it throws `Unable to get value for ValueResolver(Trees.age)`.

```
start organism Tree

  age.init = 0 count
  age.step = prior.age + 1 count

  maturity.init = 0 count
  # NO default maturity.step handler!

  # Start in "seedling" - but maturity has no handler for seedling!
  state.init = "seedling"
  # Stay in seedling forever (no transition)
  state.step = "seedling"

  inSeedling.init = 1 count
  inSeedling.step
    :if(current.state == "seedling") = 1 count
    :else = 0 count

  # Handlers only exist for juvenile and adult - NOT seedling
  start state "juvenile"
    maturity.step = current.age / 10 count
  end state

  start state "adult"
    maturity.step = 1 count
  end state

end organism
```

Intuitively I think this should maybe throw a warning, but you could also imagine plenty of situations where the user want the initial state to be vanilla and do nothing interesting (changing state by some patch-level or unconditional organism logic for eg), then having stateful behaviors trigger once they get to some state. In either case I would expect to run properly

sampottinger:
I suppose this is a little undefined right now but that makes sense. If the state is not found, it should just return the an empty handler set.

Issue #321, opened 2025-11-30
Bug - Invalid lack of initial state doesn't throw


GondekNP:
I noticed that we are actually allowed to NOT initialize state and attempt to use the conditional state handlers:

```
start organism TreeNoStateInit

  age.init = 0 count
  age.step = prior.age + 1 count

  maturity.init = 0 count

  # NO state.init! Only state.step
  state.step:if(current.maturity > 0.5 count) = "adult"

  isInState.init = 0 count
  isInState.step
    :if(current.state == "adult") = 1 count
    :else = 0 count

  start state "juvenile"
    maturity.step = current.age / 10 count
  end state

  start state "adult"
    maturity.step = 1 count
  end state

end organism
```

This throws an error at runtime (`Unable to get value for ValueResolver(current.state)`). But I'm guessing this should be caught at validation time, or failing that, at least throw a much more helpful error.

sampottinger:
Hey there! It turns out this was integer division on `current.age / 10 count`. So, I guess it is technically implemented correctly but I think our users might prefer autowidening to decimal during integer division? Will put in a PR and merge to #319 to make it easy to reverse if needed.

PR #334, opened 2026-01-13
[dev-restore] Debug / print feature

GondekNP:
This is essentially a reimplementation of #326, using a different approach. 

#326 fully abstracts export and debug logic into generics (`OutputWriter<T>`, `WriteTask<T>`, `OutputWriterFactory`, `SandboxOutputWriterFactory`, etc etc), but introduced like 1500+ lines and 11 new files to do so. Also, based on the fact that the debug logic was implemented to try and address #321, #322, and #323, there is some contamination of organism lifecycle changes and debug logic (primarily due to poor branch discipline on my part!!!). 

This branch implements the same feature of being able to use `debug()` expressions to get rich templated printouts of agent / simulation behavior as defined below. However, it does so by duplicating:
  - Queue lifecycle pattern (start/join) - ~30 lines
  - Combined routing pattern (by entity type) - ~100 lines
  - WASM sandbox delegation pattern - ~50 lines

But, in doing so, we avoid having to monkey around with working export logic and (I think) have much less abstracted code to maintain in this regard. I suspect we won't be adding any new output implementations, so I think it's a worthy tradeoff. 

___________________________________

###   Configuration (in simulation block)
```
start simulation Main
  debugFiles.patch = "file:///tmp/patch_debug.txt"
  debugFiles.organism = "file:///tmp/organism_debug.txt"
end simulation
```

### Usage (in entity blocks)
```
start patch Default
  dbg.step = debug("Patch step complete")
end patch

start organism Tree
  age.step = prior.age + 1 year
  height.step = prior.height + sample uniform from 0 m to 1 m

  dbg.step = debug("Tree age:", age, "height:", height)
end organism
```

### Expected Output
```
[Step 0, organism @ 6c1bfb54 (48.5, 23.5)] Tree age: 1 height: 0.6611
[Step 1, organism @ 6c1bfb54 (48.5, 23.5)] Tree age: 2 height: 1.2345
[Step 2, organism @ 6c1bfb54 (48.5, 23.5)] Tree age: 3 height: 1.8901
```

### DIFF
--- a/examples/guide/hello_debug_cli.josh
+++ b/examples/guide/hello_debug_cli.josh
@@ -33,8 +33,7 @@ start organism ForeverTree
   height.init = 0 meters
   height.step = prior.height + sample uniform from 0 meters to 1 meters
 
-  # dbg.step = debug("Tree", identifier, " takes a step age:", age, " with height: ", height, "at location", here.x, here.y)
-  dbg.step = debug("Tree takes a step age:", age, " with height: ", height)
+  dbg.step = debug("Tree age:", age, "height:", height)
 
 end organism
 
--- a/examples/guide/hello_debug_cli.josh
+++ b/examples/guide/hello_debug_cli.josh
@@ -1,0 +1,47 @@
+start simulation Main
+
+  grid.size = 1000 m
+  grid.low = 33.7 degrees latitude, -115.4 degrees longitude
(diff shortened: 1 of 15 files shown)


GondekNP:
@sampottinger - I'm going to merge this into `dev-restore` but when it comes time to get this into `dev` / `main` I'd appreciate your review :)

PR #338, opened 2026-01-13
[dev-restore] Lambda scope for filter expressions (port PR #314)

GondekNP:
Port the lambda scope fix from dev branch (PR #314) to resolve ValueResolver errors in collection filter expressions like:
  Trees[Tree.age > 5 years]

Previously, Tree.age could not be resolved inside the filter brackets because "Tree" was not in scope during filter evaluation.

Changes:
- LocalScope: Add defineConstantAllowShadowing() to allow temporary shadowing of parent scope variables during filter evaluation
- EventHandlerMachine: Add peek() and withLocalBinding() interface methods
- SingleThreadEventHandlerMachine: Implement peek() and withLocalBinding() with proper scope management using try-finally
- JoshDistributionVisitor: Update visitSlice() to bind the type name (e.g., "Tree") to the subject distribution before evaluating the filter expression

### DIFF
--- a/src/main/java/org/joshsim/engine/func/LocalScope.java
+++ b/src/main/java/org/joshsim/engine/func/LocalScope.java
@@ -84,6 +84,28 @@ public class LocalScope implements Scope {
   }
 
   /**
+   * Defines a constant value in the local scope, allowing shadowing of parent scope variables.
+   *
+   * <p>Unlike {@link #defineConstant}, this method only checks if the variable exists in this
+   * local scope, not in the parent scope. This allows temporarily shadowing variables from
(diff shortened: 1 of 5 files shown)


GondekNP:
@sampottinger another one I'd appreciate review before it makes it to dev. A good example:

`josh/josh-tests/conformance/core/collections/test_collections_filter_basic.josh` :

```
oldTrees.step:if(meta.stepCount >= 2 count) = Trees[Tree.age > 3 years]
```

would previously fail with `java.lang.IllegalStateException: Unable to get value for ValueResolver(Tree.age)`

PR #350, opened 2026-01-21
Rebase `dev` with `dev-restore`

GondekNP:
The dev branch got polluted with buggy fix attempts (PRs #311, #314, #315) that were interleaved with some good features, primarily developed to help diagnose the bug (debug output, timestamp / templating updates, random seed, conformance tests). The `state` bugs were actually fixed in a clean PR (#319) on main. On `dev-restore` we re-implemented the good features from scratch on top of main, rather than deal with merge conflicts and potential bugs.

| PR | Title | Risk | Impact | What Changed |
|----|-------|------|--------|--------------|
| #333 | Conformance Tests | 🟢 Low | Additive | 117 test files + JUnit infrastructure |
| #334 | Debug Output System | 🔴 High | Additive + Engine | Grammar changes to `JoshLang.g4`, new IO subsystem |
| #335 | Timestamp Templating | 🟢 Low | Additive | `{timestamp}` in `TemplateStringRenderer` (has unit tests) |
| #336 | Random Seed | 🟡 Med | Engine | `--seed` CLI flag, `SharedRandom` utility, serial execution mode |
| #337 | Test Data Generation | 🟢 Low | Additive | Gradle tasks for GeoTIFF/NetCDF/JSHD generation |
| #338 | Lambda Scope Fix | 🟡 Med | Engine | Scope resolution in filter lambdas (fresh impl, not cherry-pick) |
| #339 | Meta Synthetic Variables | 🟢 Low | Engine | `meta.stepCount`, `meta.year` in `SyntheticScope` |
| #340 | Documentation | ⚪ None | Docs only | `LanguageSpecification.md`, `llms-full.txt` |
| #341 | Boolean Distribution Count | 🟢 Low | Engine | Fix `count()` on boolean distributions in `RealizedDistribution` |
| #342 | Built-ins for 'here' | 🟢 Low | Engine | `here.attr` resolution in `ValueResolver` + `SyntheticScope` |
| #343 | Prior Attribute Value Fix | 🟡 Med | Engine | Prior value caching in `ShadowingEntity` (26 lines) |
| #344 | Compound Units | 🔴 High | Engine | Unit tracking in `CompoundUnits` + `EngineValue` (affects all arithmetic) |
| #345 | Boolean Epsilon | 🔴 High | Engine | Float comparison tolerance in `EngineValue` + `DecimalScalar` (affects all comparisons) |
| #346 | Map/Distribution Expansion | 🔴 High | Engine | Element-wise ops in `RealizedDistribution`, map strategies |
| #347 | Inner Entity Deduplication | 🟡 Med | Engine | Identity-based dedup in `InnerEntityGetter` (fixes double-unlock) |
| #348 | Boolean Distribution Units | 🟢 Low | Engine | Small fix in `RealizedDistribution` + test corrections |
| #351 | Quoted Compound Units in JSHC | 🟢 Low | Additive | Extend quoted compound unit parsing to `.jshc` config files |
| #352 | Debug Support for Editor | 🟡 Med | Additive + Editor | Wire debug output to joshsim.org web editor with debug panel |

**Risk Legend:**
- 🔴 **High** = Grammar changes or core value/comparison/distribution logic
- 🟡 **Med** = Engine changes with moderate scope
- 🟢 **Low** = Additive, isolated, or well-tested changes
- ⚪ **None** = Docs only

There are a few context documents on this branch to help make sense of these changes, the conformance tests, and what language choices remain (in the `context` folder within this branch:

| Document | Purpose |
|----------|---------|
| [`DEV_RESTORE_SUMMARY.md`](DEV_RESTORE_SUMMARY.md) | Tracks the 16 PRs (#333-#348) merged into dev-restore, what each changed, and which tests validate it |
| [`CONFORMANCE_ANALYSIS.md`](CONFORMANCE_ANALYSIS.md) | Breaks down all 117 conformance tests by category with verbatim code examples and links to test files |
| [`LANGUAGE_DESIGN_CHOICES.md`](LANGUAGE_DESIGN_CHOICES.md) | Tracks 8 design questions surfaced by testing: 5 resolved, 3 open (heterogeneous collections, dimensionless units, fractional exponents) |
| [`DOCUMENTATION_GAPS.md`](DOCUMENTATION_GAPS.md) | Spec analysis identifying patterns shown correctly but not explicitly warned against; recommendations for doc improvements |


GondekNP:
Note - lots of conflicts here, this will likely be a ceremonial PR - either we just rebase `dev-restore` on top of `dev`before merge, or we just close this and force push to dev

GondekNP:
Also - for the purpose of review, this is targeted at `main` because that is the meaningful diff (it's not particularly helpful for us to compare against the buggy dev branch).

src/main/antlr/org/joshsim/lang/antlr/JoshConfig.g4
```diff
@@ -59,4 +62,4 @@ identifier : IDENTIFIER_ ;
 
 number : NUMBER_ ;
 
-value : number (identifier | PERCENT_)? ;
+value : number (identifier | PERCENT_ | STR_)? ;
```
GondekNP, line 65: from #351 - for the purpose of accepting compound units from jshc files:
```
rainfall = 50 "mm / month"
```

src/main/antlr/org/joshsim/lang/antlr/JoshLang.g4
```diff
@@ -100,13 +101,13 @@ IDENTIFIER_: [A-Za-z][A-Za-z0-9]*;
...
 // Identifiers
-nakedIdentifier: (IDENTIFIER_|INIT_|START_|STEP_|END_|HERE_|CURRENT_|PRIOR_|STATE_|ASSERT_|PATCH_|SIMULATION_|AGENT_);
+nakedIdentifier: (IDENTIFIER_|DEBUG_|INIT_|START_|STEP_|END_|HERE_|CURRENT_|PRIOR_|STATE_|ASSERT_|PATCH_|SIMULATION_|AGENT_|ORGANISM_);
 identifier: nakedIdentifier (DOT_ (nakedIdentifier))*;
 
 // Values
 number: (SUB_|ADD_)? (FLOAT_ | INTEGER_);
 
-unitsValue: number (identifier|PERCENT_);
+unitsValue: number (identifier|PERCENT_|STR_);
```
GondekNP, line 110: Similar logic to `JoshConfig`, from #344

src/main/antlr/org/joshsim/lang/antlr/JoshLang.g4
```diff
@@ -128,6 +129,7 @@ expression: unitsValue # simpleExpression
   | operand=expression AS_ target=identifier # cast
   | FORCE_ operand=expression AS_ target=identifier # castForce
   | name=funcName LPAREN_ operand=expression RPAREN_ # singleParamFunctionCall
+  | name=funcName LPAREN_ args=expressionList RPAREN_ # variadicFunctionCall
```
GondekNP, line 132: `expressionList` was necessary for debug functions in the form:
```
debug.step = debug("age:", current.age, "height:", current.height)
```

GondekNP: from #352

src/main/java/org/joshsim/command/RunCommand.java
```diff
@@ -206,6 +214,22 @@ public Integer call() {
...
+    // Initialize shared random with seed for reproducibility
+    if (seed != null) {
+      SharedRandom.initialize(seed);
+      output.printInfo("Using random seed: " + seed);
+
+      // Force serial execution when seeded for deterministic results
+      // Parallel execution would cause non-deterministic random call ordering
+      if (!serialPatches) {
+        output.printInfo(
+            "Note: Forcing serial patch execution for reproducibility with --seed");
```
GondekNP, line 226: From #336 - a `SharedRandom` for all patches. It's documented in #336, but since our patches are streamed in parallel, I force these to run serially to avoid non-determinism in activation order. In the end our conformance tests are still using ranges, but at least we don't get random failures due to happenstance

src/main/java/org/joshsim/engine/value/type/DecimalScalar.java
```diff
@@ -128,4 +128,56 @@ protected EngineValue unsafeRaiseToPower(EngineValue other) {
...
   }
+
+  /**
+   * Epsilon value for floating-point equality comparison.
+   *
+   * <p>This value is used to determine if two decimal numbers are "equal"
+   * within acceptable precision limits. Some operations involve conversion
+   * through doubles which can introduce small rounding errors.</p>
+   */
+  private static final double EPSILON = 1e-10;
```
GondekNP: I noticed that assertions were throwing due to floating point errors - we didn't notice this because in our actual JUnit tests we allow for imprecision 😅

src/main/java/org/joshsim/engine/value/type/DoubleScalar.java
```diff
@@ -127,4 +127,57 @@ protected EngineValue unsafeRaiseToPower(EngineValue other) {
...
   }
+
+  /**
+   * Epsilon value for floating-point equality comparison.
+   *
+   * <p>This value is used to determine if two floating-point numbers are "equal"
+   * within acceptable precision limits. Standard floating-point arithmetic can
+   * produce small rounding errors (e.g., 0.4 + 0.05 = 0.45000000000000001).</p>
+   */
+  private static final double EPSILON = 1e-10;
```
GondekNP: same as DecimalScalar

src/main/java/org/joshsim/engine/value/type/RealizedDistribution.java
```diff
@@ -155,47 +157,134 @@ protected EngineValue unsafeGreaterThan(EngineValue other) {
     List<EngineValue> result = values.stream()
         .map(value -> value.greaterThan(other))
         .collect(Collectors.toCollection(ArrayList::new));
-    return new RealizedDistribution(getCaster(), result, getUnits());
+    return new RealizedDistribution(getCaster(), result, Units.EMPTY);
```
GondekNP, line 160: #348 - this was necessary because compund ANDs that use different units (`current.age > 10 and current.height < 100`) would throw due to lack of conversion between units. Really though boolean distrubution should either be unitless or logical (I opted here for unitless, but we could always go Units.LOGICAL)

src/main/java/org/joshsim/engine/value/type/RealizedDistribution.java
```diff
@@ -251,7 +340,7 @@ public EngineValue replaceUnits(Units newUnits) {
   @Override
   public EngineValue sample() {
     requireNonEmpty();
-    int index = (int) (Math.random() * values.size());
+    int index = SharedRandom.nextInt(values.size());
```
GondekNP, line 343: From #336  - for test / small example determinism

src/main/java/org/joshsim/engine/value/type/StandardVirtualDistribution.java
```diff
@@ -43,7 +42,7 @@ public StandardVirtualDistribution(EngineValueCaster caster, Units units, BigDec
    */
   @Override
   public EngineValue sample() {
-    double baseValue = random.nextGaussian();
+    double baseValue = SharedRandom.nextGaussian();
```
GondekNP, line 45: #336

src/main/java/org/joshsim/lang/bridge/InnerEntityGetter.java
```diff
@@ -81,11 +87,15 @@ public static Iterable<MutableEntity> getInnerEntities(MutableEntity target) {
...
+   * entity object appears in multiple attributes.</p>
+   *
    * @param target in which to find the inner entites
-   * @return Iterable of Entity instances found in attributes that contain entities
+   * @return Iterable of unique Entity instances found in attributes that contain entities
    */
   public static Iterable<Entity> getInnerFrozenEntities(Entity target) {
-    List<Entity> result = new ArrayList<>();
+    // Use identity-based set - same entity in multiple attributes is only returned once
+    Set<Entity> result = Collections.newSetFromMap(new IdentityHashMap<>());
```
GondekNP, line 98: `Set` vs `List` is the main change here, from #347. There were issues with entities releasing lock multiple times (`IllegalMonitorStateException`) when multiple attributes (collections for eg) contained the same entities. Probably returning duped entitie entries here would have had some side effects elsewhere??

src/main/java/org/joshsim/lang/bridge/ShadowingEntity.java
```diff
@@ -305,7 +302,9 @@ public void setAttributeValue(int index, EngineValue value) {
       resolvedCacheByIndex[index] = value;
     }
 
-    setAttributeValue(attributeName, value);
+    // NOTE: Do NOT write to inner here. Values are written to inner at endSubstep().
```
GondekNP, line 305: From #343 - Issues arose when prior was accessed one after the other:

```
start organism Tree
  height.step = prior.height + 1 m
  cover.step = prior.height / 5 m * 10 %  # Should use height BEFORE this step
end organism
``` 

In the above case prior.height would take on the new value during the same timestep, therefore we just wait to unpack and update attributes until the end of the timestep

src/main/java/org/joshsim/lang/interpret/machine/EventHandlerMachine.java
```diff
@@ -562,4 +562,51 @@ public interface EventHandlerMachine {
...
+   * Write multiple values to debug output by concatenating them.
+   *
+   * <p>Pops the specified number of values from the stack, concatenates them with spaces,
+   * and writes to the configured debug output. After writing, pushes 0 count as the result
+   * (allowing debug to be used in expressions without side effects).</p>
+   *
+   * @param argCount Number of arguments to pop from the stack.
+   * @return Reference to this machine for chaining.
+   */
+  EventHandlerMachine debugVariadic(int argCount);
```
GondekNP, line 587: This and above #334

src/main/java/org/joshsim/lang/interpret/machine/EventHandlerMachine.java
```diff
@@ -562,4 +562,51 @@ public interface EventHandlerMachine {
...
+
+  /**
+   * Peek at the top value on the stack without removing it.
+   *
+   * <p>Returns the value at the top of the stack without popping it. This is useful
+   * for inspecting values before operations that need the value to remain on the stack.</p>
+   *
+   * @return The value at the top of the stack.
+   */
+  EngineValue peek();
```
GondekNP, line 597: this and below #338 - more in `SingleThreadEventHandlerMachine`

src/main/java/org/joshsim/lang/interpret/mapping/MappingBuilder.java
```diff
@@ -111,4 +116,19 @@ public MapStrategy build(String strategyName) {
...
     };
   }
+
+  /**
+   * Normalize strategy name by converting adverb forms to base forms.
+   *
+   * @param strategyName The strategy name (may be "linear" or "linearly", etc.)
+   * @return The normalized base form ("linear", "quadratic", or "sigmoid")
+   */
+  private String normalizeStrategyName(String strategyName) {
```
GondekNP, line 126: Another language choice made that seemed harmless - #346 - in `LanguageSpecification` we use adverbs as well as linear, quadratic, sigmoid

src/main/java/org/joshsim/lang/interpret/mapping/SigmoidMapStrategy.java
```diff
@@ -7,64 +6,75 @@
...
+    double rangeLow = range.getLow().getAsDouble();
+    double rangeHigh = range.getHigh().getAsDouble();
+
+    double input = operand.getAsDouble();
+
+    // Normalize input to [-STEEPNESS, +STEEPNESS] for sigmoid
+    // This ensures the sigmoid covers most of its range (0 to 1)
+    double domainSpan = domainHigh - domainLow;
+    double t = (input - domainLow) / domainSpan; // [0, 1]
+    double x = (t * 2.0 - 1.0) * STEEPNESS; // [-STEEPNESS, +STEEPNESS]
```
GondekNP, line 64: Our original output domain was too shallow to reach the desired output domain (as in when we mapped to [0, 1], we were getting more like [.33, .66] as confirmed by conformance tests. See #346 

I don't love though this is a magic number. It's a sigmoid after all but it would be nicer in my mind to offload this to a java.math3 or something...

src/main/java/org/joshsim/lang/interpret/visitor/delegates/JoshMathematicsVisitor.java
```diff
@@ -320,4 +323,42 @@ public JoshFragment visitSingleParamFunctionCall(
...
+   * Parse a variadic function call (function with multiple arguments).
+   *
+   * <p>Currently only supports the debug() function with multiple arguments.
+   * Arguments are evaluated and concatenated with spaces for debug output.</p>
+   *
+   * @param ctx The ANTLR context from which to parse the function call.
+   * @return JoshFragment containing the function call operation parsed.
+   * @throws IllegalArgumentException if the function name is not supported for variadic calls.
+   */
+  public JoshFragment visitVariadicFunctionCall(
```
GondekNP, line 336: plumbing for #334

src/main/java/org/joshsim/lang/interpret/visitor/delegates/JoshValueVisitor.java
```diff
@@ -136,6 +137,11 @@ private EngineValue parseUnitsValue(JoshLangParser.UnitsValueContext ctx) {
     boolean hasDecimal = numberStr.contains(".");
     boolean isPercent = unitsText.equals("percent") || unitsText.equals("%");
 
+    // Handle quoted compound units like "m / yr" - strip quotes and pass to Units.of()
+    if (unitsText.startsWith("\"") && unitsText.endsWith("\"")) {
```
GondekNP, line 141: related to `JoshLang.g4` updates for #344 - this allows us to define compound units in josh files. We previously _could_ use a workaround:
```
biomass.step = prior.biomass + (current.height * (0.1 kg / 1 m))
```
Which would properly resolve .1 kg/m using dimensional analysis, but is awkward and hard to read. 

Instead we allow quoted units now:

```
biomass.step = prior.biomass + (current.height * 0.1 "kg / m")
```

However, other approaches still work. This quoting business seemed least dangerous in the sense that we already in the context of a unit, so changes were pretty narrow. I think ergonomically it might be preferable to NOT require quotes, though, but definitely leaves room for ambiguity with variables/expressions...

src/main/java/org/joshsim/lang/interpret/BridgeGetter.java
```diff
@@ -33,4 +35,27 @@ default void setBridge(EngineBridge bridge) {
...
 
+  /**
+   * Sets the debug output facade for writing debug messages.
+   *
+   * <p>This allows injecting a configured debug output facade so that debug()
+   * function calls can write to the appropriate destinations.</p>
+   *
+   * @param debugOutputFacade The debug output facade to use.
+   */
+  default void setDebugOutputFacade(CombinedDebugOutputFacade debugOutputFacade) {
```
GondekNP, line 46: more plumbing for #334

src/main/java/org/joshsim/lang/interpret/FutureBridgeGetter.java
```diff
@@ -99,6 +102,24 @@ public void setInputOutputLayer(InputOutputLayer inputOutputLayer) {
     this.inputOutputLayer = Optional.of(inputOutputLayer);
   }
 
+  /**
+   * Sets the debug output facade for debug output.
+   *
+   * @param debugOutputFacade The debug output facade to use for debug() function calls.
+   */
+  @Override
+  public void setDebugOutputFacade(CombinedDebugOutputFacade debugOutputFacade) {
```
GondekNP: again more plumbing for #334

src/main/java/org/joshsim/lang/io/CombinedDebugOutputFacade.java
```diff
@@ -0,0 +1,154 @@
...
+ * @license BSD-3-Clause
+ */
+
+package org.joshsim.lang.io;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Routes debug messages to different output facades based on entity type.
```
GondekNP, line 13: Following the pattern of CombinedExportFacade here.... per-entity-type facades

src/main/java/org/joshsim/lang/io/DebugOutputFacadeBuilder.java
```diff
@@ -0,0 +1,140 @@
...
+
+      if (value.isPresent()) {
+        try {
+          String templatePath = value.get().getAsString();
+          String path = exportFactory.getPath(templatePath);
+          OutputStreamStrategy strategy = createStrategyFromPath(path, exportFactory);
+          facades.put(entityType, new DebugOutputFacade(strategy));
+        } catch (Exception e) {
+          // Skip this entity type if strategy creation fails (e.g., in web environment)
+          // Debug output will be silently ignored for this entity type
```
GondekNP, line 82: We could be stricter about this, but I figure debugging failing silently is probably okay

src/main/java/org/joshsim/lang/io/JvmExportFacadeFactory.java
```diff
@@ -53,6 +60,7 @@ public JvmExportFacadeFactory(int replicate, TemplateStringRenderer templateRend
     this.replicate = replicate;
     this.templateRenderer = templateRenderer;
     this.minioOptions = minioOptions;
+    this.timestamp = TIMESTAMP_FORMAT.format(Instant.now().atZone(ZoneId.systemDefault()));
```
GondekNP, line 63: #335 - timestamp support for exports (crude way to avoid collisions)

src/main/java/org/joshsim/util/SharedRandom.java
```diff
@@ -0,0 +1,177 @@
...
+ *
+ * // In simulation code
+ * double value = SharedRandom.get().nextDouble();
+ *
+ * // At simulation end (cleanup)
+ * SharedRandom.clear();
+ * </pre>
+ * </p>
+ */
+public final class SharedRandom {
```
GondekNP, line 38: #336

src/main/java/org/joshsim/lang/interpret/machine/SingleThreadEventHandlerMachine.java
```diff
@@ -68,26 +74,63 @@ public class SingleThreadEventHandlerMachine implements EventHandlerMachine {
...
-        () -> new IllegalStateException("Unable to get value for " + valueResolver)
-    ));
-    return this;
+    if (value.isPresent()) {
+      memory.push(value.get());
+      return this;
+    }
+
+    // Fall back to built-in meta attributes (e.g., meta.year, meta.stepCount)
+    if (path.startsWith("meta.")) {
```
GondekNP, line 114: #339 - if normal resolution fails (meta doesn't exist), try built-ins (same logic as `here` below)

src/main/java/org/joshsim/lang/interpret/machine/SingleThreadEventHandlerMachine.java
```diff
@@ -68,26 +74,63 @@ public class SingleThreadEventHandlerMachine implements EventHandlerMachine {
...
+      String attrName = path.substring(5); // Remove "meta." prefix
+      Optional<EngineValue> builtin = getBuiltinMetaAttribute(attrName);
+      if (builtin.isPresent()) {
+        memory.push(builtin.get());
+        return this;
+      }
+    }
+
+    // Fall back to built-in here attributes (e.g., here.x, here.y)
+    if (path.startsWith("here.")) {
```
GondekNP, line 124: same above, but for #342

src/main/java/org/joshsim/lang/interpret/machine/SingleThreadEventHandlerMachine.java
```diff
@@ -531,9 +571,78 @@ public EventHandlerMachine executeSpatialQuery(ValueResolver resolver) {
   public EventHandlerMachine pushAttribute(ValueResolver resolver) {
     Entity entity = pop().getAsEntity();
     Scope scope = new EntityScope(entity);
-    EngineValue attributeValue = resolver.get(scope).orElseThrow();
-    memory.push(attributeValue);
-    return this;
+
+    // Try normal resolution first
+    Optional<EngineValue> value = resolver.get(scope);
```
GondekNP, line 576: Looking at this now, we probably should be opinionated and not allow `here` and `meta` to be overwritten

src/main/java/org/joshsim/lang/interpret/machine/SingleThreadEventHandlerMachine.java
```diff
@@ -531,9 +571,78 @@ public EventHandlerMachine executeSpatialQuery(ValueResolver resolver) {
...
+  /**
+   * Get a built-in meta attribute value if the name matches a built-in.
+   *
+   * <p>Built-in meta attributes are available without user definition per the language spec:
+   * meta.stepCount (0-based step counter), meta.year/meta.step (current step value).</p>
+   *
+   * @param name the attribute name to check.
+   * @return Optional containing the value if it's a built-in, empty otherwise.
+   */
+  private Optional<EngineValue> getBuiltinMetaAttribute(String name) {
```
GondekNP, line 604: #339 - this is the much nicer and more readable part that we can iterate on if we want to expose any other `meta` vars from bridge

src/main/java/org/joshsim/lang/interpret/machine/SingleThreadEventHandlerMachine.java
```diff
@@ -550,7 +659,7 @@ public EventHandlerMachine randUniform() {
     if (Math.abs(maxDouble - minDouble) < 1e-7) {
       doubleResult = minDouble;
     } else {
-      doubleResult = random.nextDouble(minDouble, maxDouble);
+      doubleResult = SharedRandom.nextDouble(minDouble, maxDouble);
```
GondekNP, line 662: #336

src/main/java/org/joshsim/lang/interpret/machine/SingleThreadEventHandlerMachine.java
```diff
@@ -990,4 +1183,82 @@ private boolean isQueryForPatch(ValueResolver resolver, String patchName) {
...
+  }
+
+  /**
+   * Apply a unary function to each element of a distribution.
+   *
+   * @param distribution The distribution (realized or virtual) to apply the function to.
+   * @param function The function to apply to each element.
+   * @return A new RealizedDistribution containing the transformed values.
+   */
+  private EngineValue applyToDistribution(
```
GondekNP, line 1241: All those math functions at lines 696-745 are calling this to enable element-wise math for `abs`, `ceil`, `floor`, `round`, `log10` and `ln`. I think it might be reasonable to just straight up not support _some_ of these, but I think it's intuitive to apply them elementwise

src/main/java/org/joshsim/lang/interpret/visitor/delegates/JoshDistributionVisitor.java
```diff
@@ -51,7 +55,16 @@ public JoshFragment visitSlice(JoshLangParser.SliceContext ctx) {
 
     EventHandlerAction action = (machine) -> {
       subjectAction.apply(machine);
-      selectionAction.apply(machine);
+
+      // Extract type name and bind to local scope for filter expressions
+      EngineValue subject = machine.peek();
```
GondekNP, line 60: from `EventHandlerMachine` #338. This is basically enabling:

```
start organism Tree
  age.init = 10 years
end organism

start patch Forest
  Trees.init = create 100 Tree
  oldTrees.step = Trees[Tree.age > 50 years]
end patch
```

Previously this (`oldTrees.step = Trees[Tree.age > 50 years]`)failed to resolve `Tree`, because `Tree` isn't a valid variable int the scope of the Patch (only Trees and oldTrees). 

So basically, best I understand, this is PEEKING (not popping / consuming) at the first element of Trees, getting it's langaugeType - in this case Tree, and then creates a nested scope where 'Tree' correctly resolves to the element being evaluated. Then, we go back to normal (Tree no longer in scope)

src/main/java/org/joshsim/lang/interpret/machine/SingleThreadEventHandlerMachine.java
```diff
@@ -882,6 +921,160 @@ public void pushConfigWithDefault(String name) {
...
+    CombinedDebugOutputFacade facade = debugOutputFacade.get();
+    facade.write(message, step, info.entityCategory, info.identifier, info.x, info.y);
+
+    // Push the original value back (debug is a pass-through function)
+    push(value);
+    return this;
+  }
+
+  @Override
+  public EventHandlerMachine debugVariadic(int argCount) {
```
GondekNP, line 948: come to think of it, I'm not sure we can't just treat everything variadic because even one element can be safely treated in this function (if argCount is 1)....

src/main/java/org/joshsim/engine/value/type/VirtualDistribution.java
```diff
@@ -278,4 +283,22 @@ protected EngineValue unsafeNotEqualTo(EngineValue other) {
     return realized.unsafeNotEqualTo(other);
   }
 
+  @Override
+  protected EngineValue unsafeAnd(EngineValue other) {
```
GondekNP, line 287: Note that we don't have boolean virtual distributions right now so these are probably not necessary, but were implemented to satisfy abstract class methods added by Element-wise boolean stuff (#341)

GondekNP (commented):
@sampottinger - hope it helps!!!! 😜

.github/scripts/parse_test_failures.py
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env python3
+"""Parse JUnit XML test results and output failure details as markdown."""
```
sampottinger, line 3: May want to include a "License:" entry

.github/workflows/conformance.yaml
```diff
@@ -0,0 +1,198 @@
...
+      - name: Run conformance tests
+        id: conformance
+        continue-on-error: true
+        run: ./gradlew conformanceTest
+
+      - name: Parse test results
+        if: always()
+        id: parse
+        run: |
+          # Parse JUnit XML results
```
sampottinger: This is a little bit on the longer side. Should we consider putting this in a sh file instead?

.github/workflows/conformance.yaml
```diff
@@ -0,0 +1,198 @@
...
+          echo "summary=$SUMMARY" >> $GITHUB_OUTPUT
+          echo "passed=$PASSED" >> $GITHUB_OUTPUT
+          echo "failed=$FAILED" >> $GITHUB_OUTPUT
+          echo "total=$TOTAL" >> $GITHUB_OUTPUT
+          echo "pass_rate=$PASS_RATE" >> $GITHUB_OUTPUT
+
+      - name: Generate failure report
+        if: always()
+        run: |
+          RESULTS_DIR="build/test-results/conformanceTest"
```
sampottinger: This is a little bit on the longer side. Should we consider putting this in a sh file instead?

.gitignore
```diff
@@ -53,6 +53,9 @@ landing/test_*_standalone.josh
 landing/temperature.jshd
 landing/precipitation*.jshd
 
+# Generated conformance test data - raw files are committed, JSHD files are generated
```
sampottinger: Is this comment still needed?

LanguageSpecification.md
```diff
@@ -1052,6 +1052,74 @@ end simulation
...
+
+The `debug()` function accepts any number of arguments, concatenates them with spaces, and writes to the configured output:
+
+```
+start organism Tree
+
+  age.init = 0 years
+  age.step = prior.age + 1 year
+
+  dbg.step = debug("Tree age:", age, "height:", height)
```
sampottinger, line 1088: I like this quite a lot. I wonder though about, in fitting with the other parts of the language, what you think about `debug age as "Tree age:" and height as "height:"` or similar. We don't often have C-style function calls in the grammar right now. If we do expand to include that in the future, I'm not sure if debug is the right first step. However, very open to be convinced otherwise on this point. Thanks!

LanguageSpecification.md
```diff
@@ -1052,6 +1052,74 @@ end simulation
...
+```
+
+### Output Protocols
+
+Debug output supports:
+- **file://**: Local file system (e.g., `file:///tmp/debug.txt`)
+- **minio://**: MinIO/S3 cloud storage (e.g., `minio://bucket/path/debug.txt`)
+- **stdout://**: Standard output
+
+Path templates support `{replicate}` and `{timestamp}` placeholders.
```
sampottinger, line 1121: Should it be placeholders or template tags?

LanguageSpecification.md
```diff
@@ -1371,7 +1439,7 @@ start patch Default
   Fire.step = {
     const count = 1 count if (sample uniform from 0% to 100% < 5%) else 0 count
     const new = create count of Fire
-    return prior.Fire + new;
+    return prior.Fire | new;
```
sampottinger, line 1442: Good catch!

build.gradle
```diff
@@ -110,6 +110,8 @@ test {
     events "skipped", "failed"
     exceptionFormat = "full"
   }
+  // Exclude conformance tests from default test run - they require --seed flag (Feature 4)
```
sampottinger: What is Feature 4? Also, are we skipping because of the seed or is it execution time or both?

context/DEV_RESTORE_SUMMARY.md
```diff
@@ -0,0 +1,408 @@
+# Dev-Restore Branch Summary
```
sampottinger: This was super helpful but I am not sure if it should be included in the PR or not. Please confirm!

context/CONFORMANCE_ANALYSIS.md
```diff
@@ -0,0 +1,706 @@
+# Josh Conformance Test Analysis
+
+**Status:** 111 passing (94.9%) / 6 failing (5.1%) / 117 total tests
```
sampottinger: Thank you for this! However, should this be removed before merging PR?

context/DOCUMENTATION_GAPS.md
```diff
@@ -0,0 +1,206 @@
+# Documentation Gaps and Improvement Recommendations
+
+**Last Updated:** January 21, 2026
```
sampottinger: This is useful but I am not sure this should be included in the PR itself? Should we delete prior to merge?

context/LANGUAGE_DESIGN_CHOICES.md
```diff
@@ -0,0 +1,401 @@
+# Josh Language Design Choices
+
+This document tracks language design decisions that require discussion and consensus. These are not bugs, but rather places where the language specification is ambiguous or where implementation choices affect user expectations.
```
sampottinger: Same as `DOCUMENTATION_GAPS.md`, curious if this should be excluded from the PR itself?

editor/js/main.js
```diff
@@ -208,7 +208,13 @@ class MainPresenter {
     const self = this;
     self._replicateResults = results;
     self._runPresenter.showButtons();
-    self._resultsPresenter.onComplete(self._metadata, self._replicateResults);
+
+    // Get debug store if using WASM backend
+    const debugStore = (self._currentEngineBackend && self._currentEngineBackend.getType() === "wasm")
```
sampottinger: This is great! However, I wonder about two additional thoughts for readability... First, why would currentEngineBackend ever not be set? Should confirm if this is an expected case... I actually wonder about raising an exception otherwise. We may also want to save the result of `getType() === "wasm"` to a `const isWasm` and use that in the turnery instead to try to keep this to one line? Right now it is a little long.

editor/js/model.js
```diff
@@ -692,7 +692,198 @@ class SimulationResultBuilder {
...
+    self._step = step;
+    self._entityType = entityType;
+    self._entityId = entityId;
+    self._x = x;
+    self._y = y;
+    self._content = content;
+    self._raw = raw;
+  }
+
+  getStep() { return this._step; }
```
sampottinger: For consistency, I wonder about having these as multi-line (line after opening curly and line before closing). I think it might not also be the worst idea to include JSDoc here again for consistency.

editor/js/model.js
```diff
@@ -692,7 +692,198 @@ class SimulationResultBuilder {
...
+
+  /**
+   * Parse a debug message string into a DebugMessage object.
+   *
+   * Expected format: "[Step N, entityType @ entityId (x, y)] content"
+   *
+   * @param {string} raw - The raw debug message string.
+   * @returns {?DebugMessage} Parsed message or null if parsing fails.
+   */
+  static parse(raw) {
```
sampottinger: For consistency, generally we haven't used the static option for classes and, instead, had functions at top level as it makes the imports / exports a bit cleaner. However, this is a stylistic choice so up for debate. I'm a little agnostic on return null vs throw exception but OK either way I think.

editor/js/model.js
```diff
@@ -692,7 +692,198 @@ class SimulationResultBuilder {
...
+  }
+}
+
+
+/**
+ * Stores debug messages and provides filtering capabilities.
+ */
+class DebugMessageStore {
+
+  constructor() {
```
sampottinger: Should include JSDoc if you don't mind. At the minimum, we can describe the default state of empty sets.

editor/js/model.js
```diff
@@ -692,7 +692,198 @@ class SimulationResultBuilder {
...
+      if (entityId !== null && msg.getEntityId() !== entityId) return false;
+      if (entityType !== null && msg.getEntityType() !== entityType) return false;
+      return true;
+    });
+  }
+}
+
+
 export {
+  DebugMessage,
```
sampottinger, line 698: I wonder about pulling out the two new debug classes to their own debug.js just to avoid model.js from getting a little on the larger side? What do you think? I just wonder if they are coherent enough of a unit to maybe be a separate responsibility. I could be convinced either way though.

editor/js/model.js
```diff
@@ -692,7 +692,198 @@ class SimulationResultBuilder {
...
+   * @param {?string} location - Location key to filter by, or null for all.
+   * @param {?number} step - Step number to filter by, or null for all.
+   * @param {?string} entityId - Entity ID to filter by, or null for all.
+   * @param {?string} entityType - Entity type to filter by, or null for all.
+   * @returns {Array<DebugMessage>} Filtered messages.
+   */
+  filter(location, step, entityId, entityType) {
+    const self = this;
+    return self._messages.filter((msg) => {
+      if (location !== null && msg.getLocationKey() !== location) return false;
```
sampottinger: This is a little inconsistent with other code which generally has curly for single line, especially with a more complex conditional. See around line 675. I wonder about this instead?

```
if (location !== null && msg.getLocationKey() !== location) return false;
```

To:

```
const getIsMatched = (filterVal, expected) => filterVal === null || filterVal === expected;

// ...

let isMatched = getIsMatched(location, msg.getLocation());
isMatched = isMatched && getIsMatched(step, msg.getStep());

// ...

return isMatchedOverall;
```

What do you think?

editor/js/model.js
```diff
@@ -692,7 +692,198 @@ class SimulationResultBuilder {
...
+  /**
+   * Filter messages based on criteria.
+   *
+   * @param {?string} location - Location key to filter by, or null for all.
+   * @param {?number} step - Step number to filter by, or null for all.
+   * @param {?string} entityId - Entity ID to filter by, or null for all.
+   * @param {?string} entityType - Entity type to filter by, or null for all.
+   * @returns {Array<DebugMessage>} Filtered messages.
+   */
+  filter(location, step, entityId, entityType) {
```
sampottinger: Using built-in names can sometimes lead to confusion (there is a JS native filter). What do you think about `getIsInFilter(...)`?

editor/js/wasm.js
```diff
@@ -162,6 +164,9 @@ class WasmLayer {
           const rawInput = e.data.result;
           const parsed = new OutputDatum(rawInput["target"], rawInput["attributes"]);
           self._datasetBuilder.add(parsed);
+        } else if (type === "debug") {
```
sampottinger, line 167: There's quite a lot of branching here now. I wonder if we should use an object like:

```
const strategy = {
  // ...
  "debug": () -> {
    // ...
  },
}[type];

strategy();
```

Could be convinced either way... it's on the edge.

editor/js/wire.js
```diff
@@ -123,6 +123,9 @@ class ResponseReader {
         const rawInput = intermediate["datum"];
         const parsed = new OutputDatum(rawInput["target"], rawInput["attributes"]);
         self._replicateReducer.get(intermediate["replicate"]).add(parsed);
+      } else if (intermediate["type"] === "debug") {
```
sampottinger, line 126: Again kinda on the edge but there is enough branching we might want to use an object to get a strategy function like in my last suggestion. Open to suggestion though.

editor/style/style.css
```diff
@@ -632,4 +632,156 @@ main {
   66.67% { border-color: #D0D0D0; }          /* Hold for 2 seconds */
   83.33% { border-color: #303030; }          /* Flash to dark (0.5s) */
   100% { border-color: #D0D0D0; }            /* Fade back (0.5s) */
+}
+
+/* Debug Panel Styles */
+#debug-panel {
```
sampottinger: I think this might be inconsistent with prior styling and some convention. Typically we go most broad to most general (tags, classes, ids, media queries(tags, classes, ids)) and then sort alphabetically within those categories.

examples/guide/hello_debug.josh
```diff
@@ -0,0 +1,43 @@
...
+
+start organism ForeverTree
+
+  age.init = 0 year
+  age.step = prior.age + 1 year
+
+  height.init = 0 meters
+  height.step = prior.height + sample uniform from 0 meters to 1 meters
+
+  dbg.step = debug("age:", current.age, "height:", current.height)
```
sampottinger, line 33: Nice. I wonder about `dbg` vs `debug` but if it makes the grammar easier I think that's ok.

editor/js/wasm.worker.js
```diff
@@ -18,11 +18,21 @@ function reportStepComplete(stepCount) {
...
  * Parses a data string from MemoryWriteStrategy and reports the parsed data to the main thread.
- * 
+ *
  * @param {string} source - The formatted string containing target name and key-value pairs.
  */
 function reportData(source) {
-  const datum = parseDatum(source);
+  const trimmed = source.trim();
+
+  // Check for debug messages first - they have a different format
```
sampottinger, line 27: I wonder if this might be better in a supporting function like `getIsDebug` with this information in a JSDoc comment for that function. That could offer some additional testability too potentially.

josh-tests/conformance/temporal/queries/test_temporal_queries_meta.josh
```diff
@@ -0,0 +1,122 @@
+# @category: temporal
```
sampottinger, line 1: Humm this is nice and, if we haven't yet, we should formalize what these tags mean somewhere in our docs.

llms-full.txt
```diff
@@ -523,6 +523,30 @@ export.variable.event = expression
 exportFiles.type = "output_path"
 ```
 
+### Debug Output
```
sampottinger, line 526: This looks great. Just flagging this might need an update if we change the syntax as mentioned earlier.

src/main/antlr/org/joshsim/lang/antlr/JoshConfig.g4
sampottinger: Sounds good. That said, I wouldn't be offended by a unitsIdentifier rule that allows for space and slash if we want to further constrain. LMK what you think!

src/main/antlr/org/joshsim/lang/antlr/JoshLang.g4
sampottinger: See earlier comment!

src/main/antlr/org/joshsim/lang/antlr/JoshLang.g4
sampottinger: See earlier comment about parentheses though I could be convinced either way. LMK what you think! I guess I am reminded now that we do have function calls but I still wonder about the alternative `debug var as "description" [and other as "other"]*` but curious your thoughts.

src/main/java/org/joshsim/JoshSimFacadeUtil.java
```diff
@@ -145,6 +162,7 @@ public static void runSimulation(EngineValueFactory valueFactory,
     }
 
     exportFacade.join();
+    debugFacade.join();
```
sampottinger, line 165: Nice

src/main/java/org/joshsim/command/RunCommand.java
```diff
@@ -159,6 +160,13 @@ public class RunCommand implements Callable<Integer> {
   )
   private boolean uploadData = false;
 
+  @Option(
+      names = "--seed",
+      description = "Random seed for reproducible simulations. If specified, all random "
+                  + "operations will use this seed to produce deterministic results."
+  )
+  private Long seed = null;
```
sampottinger, line 168: Very nice

src/main/java/org/joshsim/engine/func/LocalScope.java
```diff
@@ -83,6 +83,28 @@ public void defineConstant(String name, EngineValue value) {
...
+   * enclosing scopes, which is useful for filter expressions where the type name needs to
+   * be bound to the subject distribution.</p>
+   *
+   * @param name The name of the constant to define within this LocalScope.
+   * @param value The value to associate with the constant within this LocalScope.
+   * @throws RuntimeException if a variable with the given name already exists in this local scope.
+   */
+  public void defineConstantAllowShadowing(String name, EngineValue value) {
+    if (localValues.containsKey(name)) {
+      String message =
```
sampottinger: Is this consistent with prior code? I would have expected:

```
String message = String.format(
    "...",
    ...
);
```

src/main/java/org/joshsim/engine/value/type/DecimalScalar.java
sampottinger: Good catch!

src/main/java/org/joshsim/engine/value/type/DoubleScalar.java
sampottinger: I like it. I wonder if we should move to like `PrecisionUtil` even if it only contains a constant at this point?

src/main/java/org/joshsim/engine/value/type/DecimalScalar.java
```diff
@@ -128,4 +128,56 @@ protected EngineValue unsafeRaiseToPower(EngineValue other) {
...
+   * Compares this DecimalScalar to another EngineValue for inequality using epsilon tolerance.
+   *
+   * <p>This is the logical negation of {@link #unsafeEqualTo(EngineValue)}.</p>
+   *
+   * @param other the other EngineValue to compare with.
+   * @return a BooleanScalar indicating whether the values are not equal within tolerance.
+   */
+  @Override
+  protected EngineValue unsafeNotEqualTo(EngineValue other) {
+    double thisVal = getAsDouble();
```
sampottinger: If using as a double, this could be shared across the two numeric classes in a `PrecisionUtil` class?

src/main/java/org/joshsim/engine/value/type/RealizedDistribution.java
sampottinger: I agree with your choice!

src/main/java/org/joshsim/engine/value/type/RealizedDistribution.java
```diff
@@ -155,47 +157,134 @@ protected EngineValue unsafeGreaterThan(EngineValue other) {
...
   protected EngineValue unsafeNotEqualTo(EngineValue other) {
     List<EngineValue> result = values.stream()
         .map(value -> value.notEqualTo(other))
         .collect(Collectors.toCollection(ArrayList::new));
-    return new RealizedDistribution(getCaster(), result, getUnits());
+    return new RealizedDistribution(getCaster(), result, Units.EMPTY);
+  }
+
+  @Override
+  protected EngineValue unsafeAnd(EngineValue other) {
```
sampottinger, line 204: I think this is fine just noticing the asymmetry between the branches in using the Stream API. Not a strong feeling to make them the same but curious your thoughts.

src/main/java/org/joshsim/engine/value/type/RealizedDistribution.java
```diff
@@ -155,47 +157,134 @@ protected EngineValue unsafeGreaterThan(EngineValue other) {
...
+            + values.size() + " vs " + otherValues.size());
+      }
+      List<EngineValue> result = new ArrayList<>(values.size());
+      for (int i = 0; i < values.size(); i++) {
+        boolean andResult = values.get(i).getAsBoolean() && otherValues.get(i).getAsBoolean();
+        result.add(new BooleanScalar(getCaster(), andResult, Units.EMPTY));
+      }
+      return new RealizedDistribution(getCaster(), result, Units.EMPTY);
+    } else {
+      // AND each element with a scalar
```
sampottinger, line 220: Instead of the inline comments I wonder if we want to save the conditional result to a local variable `isDistributionElementWiseAnd`.

sampottinger, line 220: Instead of the inline comments I wonder if we want to save the conditional result to a local variable `isDistributionElementWiseAnd`.

src/main/java/org/joshsim/engine/value/type/RealizedDistribution.java
```diff
@@ -155,47 +157,134 @@ protected EngineValue unsafeGreaterThan(EngineValue other) {
...
+            return new BooleanScalar(getCaster(), andResult, Units.EMPTY);
+          })
+          .collect(Collectors.toCollection(ArrayList::new));
+      return new RealizedDistribution(getCaster(), result, Units.EMPTY);
+    }
+  }
+
+  @Override
+  protected EngineValue unsafeOr(EngineValue other) {
+    if (other.getLanguageType().isDistribution()) {
```
sampottinger, line 234: Same as above.

src/main/java/org/joshsim/engine/value/type/RealizedDistribution.java
```diff
@@ -155,47 +157,134 @@ protected EngineValue unsafeGreaterThan(EngineValue other) {
...
+            return new BooleanScalar(getCaster(), orResult, Units.EMPTY);
+          })
+          .collect(Collectors.toCollection(ArrayList::new));
+      return new RealizedDistribution(getCaster(), result, Units.EMPTY);
+    }
+  }
+
+  @Override
+  protected EngineValue unsafeXor(EngineValue other) {
+    if (other.getLanguageType().isDistribution()) {
```
sampottinger, line 263: Same as above

src/main/java/org/joshsim/engine/value/type/RealizedDistribution.java
```diff
@@ -155,47 +157,134 @@ protected EngineValue unsafeGreaterThan(EngineValue other) {
...
+            + values.size() + " vs " + otherValues.size());
+      }
+      List<EngineValue> result = new ArrayList<>(values.size());
+      for (int i = 0; i < values.size(); i++) {
+        boolean xorResult = values.get(i).getAsBoolean() ^ otherValues.get(i).getAsBoolean();
+        result.add(new BooleanScalar(getCaster(), xorResult, Units.EMPTY));
+      }
+      return new RealizedDistribution(getCaster(), result, Units.EMPTY);
+    } else {
+      // XOR each element with a scalar
```
sampottinger, line 278: If this structure is so commonly repeated across the logical operations, I wonder if we use a strategy to unify the logic? What do you think?

src/main/java/org/joshsim/engine/value/type/RealizedDistribution.java
```diff
@@ -265,6 +354,16 @@ public Optional<Integer> getSize() {
     return Optional.of(values.size());
   }
 
+  @Override
+  public long getCount() {
+    if (getLanguageType().getRootType().equals("boolean")) {
+      return values.stream()
+          .filter(EngineValue::getAsBoolean)
```
sampottinger, line 361: Humm I'm guessing the filter is needed but just curious why? We are getting size of distribution not number that are true right? Apologies probably not seeing something.

src/main/java/org/joshsim/lang/bridge/InnerEntityGetter.java
```diff
@@ -33,11 +34,16 @@ public class InnerEntityGetter {
    * that contain entities. Uses direct iteration instead of streams for better performance
    * in this hot path (called in every substep for every entity).</p>
    *
+   * <p>Uses identity-based Set to ensure each entity is returned only once, even if the same
```
sampottinger, line 37: Good catch

src/main/java/org/joshsim/lang/bridge/InnerEntityGetter.java
sampottinger: Should think more about this case but I think this is right. Shouldn't slow down PR.

src/main/java/org/joshsim/lang/bridge/ShadowingEntity.java
```diff
@@ -280,20 +280,17 @@ public void setAttributeValue(String name, EngineValue value) {
       }
     }
 
-    inner.setAttributeValue(name, value);
+    // NOTE: Do NOT write to inner here. Values are written to inner at endSubstep().
```
sampottinger, line 283: I think these are good ideas to capture but I feel a little weird about putting them inline here. I wonder if there is a more general statement we can make about this behavior in the JavaDoc? What do you think?

src/main/java/org/joshsim/lang/bridge/ShadowingEntity.java
sampottinger: Interesting. I guess I would have expected it to be marked as resolved and its cached value used. I'm guessing what you did here is right but I'm trying to see why it would evaluate twice in prior implementation. Did you learn more about that or am I thinking about this maybe from the wrong angle?

src/main/java/org/joshsim/lang/bridge/ShadowingEntity.java
```diff
@@ -759,6 +758,17 @@ public void startSubstep(String name) {
   public void endSubstep() {
     InnerEntityGetter.getInnerEntities(this).forEach((x) -> x.endSubstep());
 
+    // Copy resolved values from cache to inner entity.
```
sampottinger, line 761: This is great but I wonder about moving it to a supporting private method and moving the inline comments to JavaDoc? What do you think?

src/main/java/org/joshsim/lang/interpret/FutureBridgeGetter.java
```diff
@@ -99,6 +102,24 @@ public void setInputOutputLayer(InputOutputLayer inputOutputLayer) {
...
+    if (this.debugOutputFacade.isPresent()) {
+      throw new IllegalStateException("Debug output facade already set.");
+    }
+    this.debugOutputFacade = Optional.of(debugOutputFacade);
+  }
+
+  @Override
+  public Optional<CombinedDebugOutputFacade> getDebugOutputFacade() {
+    return this.debugOutputFacade;
```
sampottinger: Humm I think we have been using `this.` to avoid shadowing in constructor but otherwise omitting (https://github.com/SchmidtDSE/josh/blob/dev-restore/src/main/java/org/joshsim/wire/WireParseResult.java#L63). Should we remove `this.` here?

src/main/java/org/joshsim/lang/interpret/ValueResolver.java
```diff
@@ -170,6 +170,15 @@ private Optional<EngineValue> tryIntegerLookup(EntityScope entityScope) {
     return entityScope.getOptional(index);
   }
 
+  /**
+   * Get the path being resolved.
+   *
+   * @return the dot-separated path string.
+   */
+  public String getPath() {
+    return path;
```
sampottinger, line 179: Nice - surprised we didn't need this already.

src/main/java/org/joshsim/lang/interpret/machine/EventHandlerMachine.java
sampottinger: Nice

src/main/java/org/joshsim/lang/interpret/mapping/MappingBuilder.java
sampottinger: Nifty

src/main/java/org/joshsim/lang/interpret/visitor/JoshConfigParserVisitor.java
```diff
@@ -96,6 +96,11 @@ public JshcFragment visitValue(JoshConfigParser.ValueContext ctx) {
       if (unitsText.equals("percent")) {
         isPercent = true;
       }
+    } else if (ctx.STR_() != null) {
```
sampottinger, line 99: I think this is OK though may be another vote for having an indentifierOrUnit rule that allows for identifiers with "\s?\/\s?" Just to make sure this string manipulation is consistent. LMK what you think of the earlier comment though. This might stay otherwise.

src/main/java/org/joshsim/lang/interpret/visitor/delegates/JoshDistributionVisitor.java
sampottinger: Nice

src/main/java/org/joshsim/lang/interpret/visitor/delegates/JoshMathematicsVisitor.java
sampottinger: Looks good!

src/main/java/org/joshsim/lang/interpret/visitor/delegates/JoshValueVisitor.java
sampottinger: Interesting. If I understand correctly, this is just additive so both statements would work correctly now? Again might be worth considering that identifierOrUnit but curious your thoughts.

src/main/java/org/joshsim/lang/io/CombinedDebugOutputFacade.java
```diff
@@ -0,0 +1,154 @@
...
+   *
+   * @param message The debug message to write.
+   * @param step The current simulation step number.
+   * @param entityType The entity type for routing.
+   * @param identifier Unique identifier for the entity (hex hash).
+   * @param x X coordinate of the entity's location.
+   * @param y Y coordinate of the entity's location.
+   */
+  public void write(String message, long step, String entityType,
+                    String identifier, double x, double y) {
```
sampottinger: I think we were using consistent indent not aligned to opening parentheses?

src/main/java/org/joshsim/lang/io/CombinedDebugOutputFacade.java
sampottinger: Nice! The lookup by entity name could end up being slow if on critical but if we are debugging we are already accepting some performance hit so I think this is OK. I think this is dormant unless a user adds debug statements?

src/main/java/org/joshsim/lang/io/DebugOutputFacade.java
```diff
@@ -0,0 +1,168 @@
...
+   *
+   * @param message The debug message to write.
+   * @param step The current simulation step number.
+   * @param entityType The type of entity generating the message (e.g., "patch", "agent").
+   * @param identifier Unique identifier for the entity (hex hash).
+   * @param x X coordinate of the entity's location.
+   * @param y Y coordinate of the entity's location.
+   */
+  public void write(String message, long step, String entityType,
+                    String identifier, double x, double y) {
```
sampottinger: See earlier comment and prior code like https://github.com/SchmidtDSE/josh/blob/dev-restore/src/main/java/org/joshsim/cloud/JoshSimServer.java#L56

src/main/java/org/joshsim/lang/io/DebugOutputFacade.java
```diff
@@ -0,0 +1,168 @@
...
+    DebugMessage debugMessage = new DebugMessage(message, step, entityType, identifier, x, y);
+    queueService.add(debugMessage);
+  }
+
+  /**
+   * Internal record to hold debug message data for queue processing.
+   */
+  private record DebugMessage(String message, long step, String entityType,
```
sampottinger: See above

src/main/java/org/joshsim/lang/io/DebugOutputFacade.java
```diff
@@ -0,0 +1,168 @@
...
+                              String identifier, double x, double y) {}
+
+  /**
+   * Internal callback that handles writing debug messages to the output stream.
+   *
+   * <p>This follows the same pattern as the InnerWriter in CsvExportFacade.</p>
+   */
+  private static class InnerWriter implements QueueServiceCallback {
```
sampottinger, line 113: I am OK with this but, if moving this to a top level public class provides additional testing, I'm in favor of that. LMK what you think!

src/main/java/org/joshsim/lang/io/DebugOutputFacade.java
```diff
@@ -0,0 +1,168 @@
...
+
+    @Override
+    public void onTask(Optional<Object> taskMaybe) {
+      if (taskMaybe.isEmpty()) {
+        writer.flush();
+        return;
+      }
+
+      DebugMessage msg = (DebugMessage) taskMaybe.get();
+      String formatted = String.format("[Step %d, %s @ %s (%.1f, %.1f)] %s",
```
sampottinger: I wonder if this might be better as one parameter per line given length? This is good but there's a lot going on in here.

src/main/java/org/joshsim/lang/io/DebugOutputFacadeBuilder.java
```diff
@@ -0,0 +1,140 @@
...
+   *
+   * <p>Uses the ExportFacadeFactory to create OutputStreamStrategy instances, allowing
+   * debug output to use the same infrastructure (file, MinIO) as export output.</p>
+   *
+   * @param simEntity The simulation entity containing debugFiles.* attributes.
+   * @param exportFactory The export factory to use for creating output stream strategies.
+   * @return A CombinedDebugOutputFacade configured based on the entity's attributes.
+   */
+  public static CombinedDebugOutputFacade build(MutableEntity simEntity,
+                                                 ExportFacadeFactory exportFactory) {
```
sampottinger: See above and https://github.com/SchmidtDSE/josh/blob/dev-restore/src/main/java/org/joshsim/cloud/JoshSimServer.java#L56

src/main/java/org/joshsim/lang/io/DebugOutputFacadeBuilder.java
sampottinger: Yeah I wonder too. Not sure if it should emit printf or just capture and quietly suppress is ok. Up to either answer I think in this case!

src/main/java/org/joshsim/lang/io/JvmExportFacadeFactory.java
sampottinger, line 63: I think this is OK but we might want to move to a supporting private method to document purpose.

src/main/java/org/joshsim/lang/io/JvmExportFacadeFactory.java
```diff
@@ -90,6 +98,7 @@ public JvmExportFacadeFactory(int replicate, PatchBuilderExtents extents, BigDec
...
```
sampottinger, line 100: Same as above.

src/test/java/org/joshsim/conformance/JoshConformanceTest.java
```diff
@@ -0,0 +1,183 @@
...
+
+
+/**
+ * Parameterized JUnit test that discovers and runs all Josh conformance tests.
+ *
+ * <p>Each .josh file with assertions is a self-validating test.
+ * This runner simply executes them and checks the exit code.</p>
+ */
+@ExtendWith(PerformanceTracker.class)
+class JoshConformanceTest {
```
sampottinger, line 30: Nifty. My only thing is what do you think about adding a `README.md` to that new directory just to describe that we are doing some discovery in there through this test? Just a little worried that the discovery mechanism might not be obvious at first glance. WDYT?

src/test/java/org/joshsim/conformance/PerformanceTracker.java
```diff
@@ -0,0 +1,155 @@
...
+        .mapToDouble(d -> Math.pow(d - baselineMean, 2))
+        .average()
+        .orElse(0);
+    double stdDev = Math.sqrt(variance);
+
+    double slowdown = (currentMs - baselineMean) / baselineMean;
+    double zscore = stdDev > 0 ? (currentMs - baselineMean) / stdDev : 0;
+
+    if (slowdown > REGRESSION_THRESHOLD || zscore > Z_SCORE_THRESHOLD) {
+      System.err.println(String.format(
```
sampottinger, line 151: I get a little nervous around including emojis in live source outside HTML. There can be some weird stuff that happens with some editors and it can cause accessibility issues. That said, I don't think it's bad that we included these. Maybe it makes sense to pull these strings up into multiline string constants top of file? https://blog.jetbrains.com/idea/2025/06/text-blocks-in-java-perfect-for-multiline-strings/

src/test/java/org/joshsim/conformance/PerformanceTracker.java
```diff
@@ -0,0 +1,155 @@
...
+          + "   Current: %dms\n"
+          + "   Baseline: %.0fms (±%.0fms)\n"
+          + "   Slowdown: %.1f%% (threshold: %.0f%%)\n"
+          + "   Z-score: %.2f (threshold: %.1f)\n",
+          testName, currentMs, baselineMean, stdDev,
+          slowdown * 100, REGRESSION_THRESHOLD * 100, zscore, Z_SCORE_THRESHOLD
+      ));
+    }
+
+    if (slowdown < -IMPROVEMENT_THRESHOLD && recentRuns.size() >= BASELINE_WINDOW) {
```
sampottinger, line 163: See above

src/test/java/org/joshsim/conformance/TestDataGenerator.java
```diff
@@ -0,0 +1,395 @@
...
+ * <p>Usage:
+ * <pre>
+ * // Standalone execution
+ * java org.joshsim.conformance.TestDataGenerator
+ *
+ * // Via Gradle
+ * ./gradlew generateTestData
+ * </pre>
+ */
+public class TestDataGenerator {
```
sampottinger, line 61: I think this looks great. An opinion loosely held, I wonder if we should do some more functional decomposition though. Some of the methods are quite long. We could move some inline comments into private method JavaDoc as well.

src/test/java/org/joshsim/conformance/TestMetadata.java
```diff
@@ -0,0 +1,105 @@
...
+   * Constructs a TestMetadata instance with the specified fields.
+   *
+   * @param category The test category (e.g., "lifecycle", "types", "spatial")
+   * @param subcategory The test subcategory (e.g., "events", "conversions")
+   * @param priority The test priority (e.g., "critical", "high", "medium", "low")
+   * @param issue The related issue number (e.g., "#123")
+   * @param description The human-readable test description
+   */
+  TestMetadata(String category, String subcategory, String priority,
+               String issue, String description) {
```
sampottinger: Indent possibly should follow closer to existing code like https://github.com/SchmidtDSE/josh/blob/dev-restore/src/main/java/org/joshsim/cloud/JoshSimServer.java#L57.

src/test/java/org/joshsim/conformance/TestMetadataTest.java
```diff
@@ -0,0 +1,279 @@
...
+ */
+class TestMetadataTest {
+
+  @TempDir
+  Path tempDir;
+
+  @Test
+  void testParseCompleteMetadata() throws Exception {
+    String content = """
+        # @category: lifecycle
```
sampottinger, line 29: Nice. I think I see what's up now. I wonder if maybe that README in the new Josh tests directory could maybe serve to document the tags expected / optional.

src/test/java/org/joshsim/engine/value/type/DoubleScalarTest.java
```diff
@@ -165,4 +166,80 @@ void testScaleAndPrecision() {
     assertEquals(123.456789, scalar.getAsDouble(), 0.000001);
     assertEquals("123.456789", scalar.getAsString());
   }
+
+  @Test
+  void testEqualToWithArithmeticResult() {
+    // This is the critical test case: 0.4 + 0.05 should equal 0.45
```
sampottinger, line 172: Unsure if the comment is helpful or if obvious given the code after.

sampottinger (dismissed):
Thanks @GondekNP! I think this is great. I think I had some minor tweaks though it might be good to have conversation around the debug syntax and some additional documentation of the conformance tests through an additional README. Obviously everything up for discussion. Please let me know what you think! Thanks for all of this! Sending on a Sunday because I may have some publications responsibilities on Monday. no rush in response!

src/main/antlr/org/joshsim/lang/antlr/JoshConfig.g4
GondekNP: If I'm understanding correctly, I think I'd prefer that - I was reticent to make any destructive changes to unit parsing (mostly because, what if someone was doing something like "10 meters / SCALING_FACTOR" or something like that). But if you see a safe way to go about it, I prefer not having these quotes because I don't think it's particularly intuitive why they'd be needed from the user perspective.

.github/workflows/conformance.yaml
GondekNP: Yeah I agreed, ugly stuff I'd prefer in a bash

src/main/antlr/org/joshsim/lang/antlr/JoshLang.g4
GondekNP: I think I might need some more explanation here 😅 

Meaning, we would have some default debug parsing and just have the debug logic link up that string with the passed var?

I think the way this is set up us much more akin to console.log() js style, which seemed most flexible for the way people like to debug, but I could be convinced (and am likely not following exactly anyways)

src/main/java/org/joshsim/engine/value/type/DoubleScalar.java
GondekNP: You mean, a general `PrecisionUtil` class for these types of operations borrowed by other implementations of scalars and such?

src/main/java/org/joshsim/lang/bridge/ShadowingEntity.java
GondekNP: I'm not 100% sure on the stack trace that gets here (I tested this condition with a breakpoint and a simple josh file that queries the same prior multiple times in a step). However my understanding is that the underlying array (`indexArray[index]`) would then be populated by the new, current value DURING the step and not after as we intended (so the current value essentially became prior). I think this (and the changes to InnerEntityGetter) would be good to tease apart a bit more because I don't fully understand the root cause at this point

src/main/java/org/joshsim/lang/interpret/machine/SingleThreadEventHandlerMachine.java
```diff
@@ -990,4 +1183,82 @@ private boolean isQueryForPatch(ValueResolver resolver, String patchName) {
     return resolver.toString().contains("ValueResolver(" + patchName + ")");
   }
 
+  @Override
+  public EngineValue peek() {
```
GondekNP, line 1187: peek() here is needed for lambda scope resolution - see `JoshDistrubutionVisitor

src/main/java/org/joshsim/lang/interpret/mapping/SigmoidMapStrategy.java
GondekNP: @sampottinger would love some feedback on whether this is reasonable / if there is a more out of the box approach we can take here. I don't love this STEEPNESS magic number, but also it seems to reproduce a sigmoid that makes sense and has a reasonable output domain

src/main/java/org/joshsim/lang/interpret/visitor/delegates/JoshValueVisitor.java
GondekNP: Yes - related to your comment on the .g4, this is additive and the old behavior stands. Basically this is a way to force the parser to interpret the entire thing and pass to Units.of() which makes everything work nice nice with dimensional analysis and such. I think there may be edge cases where we would get weird behavior without doing this - for eg dividing a number with unit by a variable name. But I am not 100% sure if I am overthinking it or if we have a more robust way to account for that situation.

src/main/java/org/joshsim/lang/io/CombinedDebugOutputFacade.java
GondekNP: Yes this is dormant unless a user explicity adds the debug output path to their josh file. In that case none of this is instantiated. would be good to confirm with some testing with big simulations though :)

.gitignore
GondekNP: Whoops nope

build.gradle
GondekNP: Also smells from context documents

LanguageSpecification.md
GondekNP: I don't love the readability of `debug age as "Tree age:" and height as "height:"` honestly. But I think this is worth a discussion. In a perfect perfect world I would prefer we have some f-string implementation like that of python, both because we already do that in our template strings for export paths (our users in theory already understand the concept) and because it's extremely clean and readable in my opinion. That or something concatenating strings with a `+`. We should discuss!

context/DEV_RESTORE_SUMMARY.md
GondekNP: Yeah agreed - I just left these in here to have some archeology on what this branch is / help on what you and I should touch base on before merging. But I think we delete this `context` folder entirely

context/CONFORMANCE_ANALYSIS.md
GondekNP: See above - delete before merge!

context/DOCUMENTATION_GAPS.md
GondekNP: See above - delete before merge!

context/LANGUAGE_DESIGN_CHOICES.md
GondekNP: See above - delete before merge!

editor/js/main.js
GondekNP: Yeah I agree - I think this was claude being overly conservative in trying to handle edge cases, I caught a few of these elsewhere (basically handlers for cases we would never expect to reach or would probably just prefer throw if so). I think this is safe to remove (and agreed on packing `isWasm`)

editor/js/model.js
GondekNP: Yeah I agree - I kind of think they are also not mission critical (by definition) so I prefer them not muddying this already enormous class

src/main/java/org/joshsim/lang/bridge/ShadowingEntity.java
GondekNP: I think this can be removed, it's basically holdover from debugging earlier. Assuming of course we feel good about the implementation after some stress testing

src/main/java/org/joshsim/lang/interpret/FutureBridgeGetter.java
GondekNP: Oh yes whoops, I remember this conversation now from the deep early days of josh :stuck_out_tongue_closed_eyes:

GondekNP:
With https://github.com/SchmidtDSE/josh/tree/214d4b214f518a919eba17ce1823260101def1fe, we force push `dev-restore` to dev. And dev is thus, :sparkle: restored! Now we close this ceremonial branch.

PR #366, opened 2026-04-07
Pin CI/CD actions to verified SHA hashes

sampottinger:
All pined to before the most recent worm with some verification that the commit is valid. This should bring back CI / CD.


GondekNP:
LGTM!!!

PR #370, opened 2026-04-09
Feat / simplify grid route

GondekNP:
This PR resolves a few spatial query bugs resultant from a couple of translation errors between Earth and Grid space at build time for a `josh` model. 

Fundamentally, `josh` doesn't run simulations in geographic space at runtime. While this was originally a concession @sampottinger and I made when trying to get `josh` to run inside of `wasm` (where geo dependencies are a nightmare to deal with), it has some runtime upside as well - by taking in geographic information at `build` time, constructing the grid in `index` terms (what we refer to as 'Grid Space') and aligning external data to these indices ahead of time, we avoid having to do any complicated geo operations during runtime since they are pre-computed. In the case of external data (`jshds`) this means that our preprocessed data are large and have a lot of duplicated values (when the grid cell size is much smaller than the external data cells for eg), and means that at runtime we have high memory demands, but for those tradeoffs we gain speed and enable the model development / evaluation process to run faster in theory. 

However, we observed some flaky and inconsistent behavior with respect to spatial queries - this is because, at the time of converting from Earth space to Grid Space, we define grid cells in terms of index units (a cell at 0.5, 0.5 w/ width 1 cell, a cell at 1.5, 0.5 w/ width 1 cell, etc) but _lose crs metadata_. For the purpose on non-spatial operations this is okay because at this point, `jshd`s already have resolved external data to the correct cell.

  ...BUT, when we construct a PatchSpatialIndex, we need to convert within R meters radial query radii from meters into grid cell units. The conversion is radiusInGridCells = diameter / (2 *  cellSizeInMeters). Before this PR, cellSizeInMeters was read from the grid cell's width — which, after count-space conversion, is 1 (index units), not the original meter value (e.g., 30m). The result is that radiusInGridCells = 60 / (2 * 1) = 30 instead of 60 / (2 * 30) = 1. 
                                                            
Bug fixes:
- Thread originalCellSizeMeters through GridCrsDefinition → Replicate → TimeStep → PatchSpatialIndex
so the meter value survives count-space conversion                                                  
- Reject unitless within distances at runtime with an actionable error — previously a unitless
neighborRadius = 2 was silently relabeled as "2 meters" via NoopConversion                    
- Refactor GridCrsDefinition API — replace the ambiguous getCellSize() (which returned index units   
but was labeled as meters) with explicit getIndexCellSize() and getOriginalCellSizeMeters()       
																								   
Runtime simplification:                                                                  
										  
The Earth-space runtime path (EarthGeometryFactory, reached via `--crs` on run) is unused by any production model, every simulation runs through count-space after preprocessing external data and basically discarding spatial information (storing only GridCrsDefinition). As mentioned above, this caused some issues when we we improperly interpreted the `width` of our `GridShape` as meters when it was really `count` - but after doing that refactor (`bug/spatial_within`) we ended up propogating `GridCrsDefinition` all the way down to frozen Patches (`GridFromSimFactory` to `PatchSet` to `Replicate` to `TimeStep` to `PatchSpatialIndex`), basically just to carry the original cell size in meters down - which is maybe a bit overkill, but it's not replicated much and now becomes the source of truth for the `josh` runtime knowing how to back-convert to real space from GridSpace, if necessary (might be wise to propogate this through spatial exports for eg).

- Removes `--crs` from RunCommand / JoshSimWorkerHandler — always uses GridGeometryFactory at runtime.  `--crs` remains on PreprocessCommand where it's needed.
- Breaks the engine → geo reverse dependency — removes getOnEarth() from EngineGeometry interface and EarthShape import from engine core. This is something we tried to avoid initially but became necessary due to our architectural choices, but now is no longer needed with the pure prepreocess strategy
- Removes deprecated getCellSize() from GridCrsDefinition

### DIFF
--- a/src/main/java/org/joshsim/cloud/JoshSimWorkerHandler.java
+++ b/src/main/java/org/joshsim/cloud/JoshSimWorkerHandler.java
@@ -20,7 +20,7 @@ import org.apache.sis.referencing.CRS;
 import org.joshsim.JoshSimFacadeUtil;
 import org.joshsim.engine.geometry.EngineGeometryFactory;
 import org.joshsim.engine.geometry.grid.GridGeometryFactory;
-import org.joshsim.engine.value.engine.EngineValueFactory;
+import org.joshsim.engine.value.engine.ValueSupportFactory;
 import org.joshsim.geo.geometry.EarthGeometryFactory;
 import org.joshsim.lang.interpret.JoshProgram;
 import org.joshsim.lang.io.InputOutputLayer;
@@ -171,7 +171,7 @@ public class JoshSimWorkerHandler implements HttpHandler {
     }
 
     InputOutputLayer inputOutputLayer = getLayer(httpServerExchange, externalData);
-    EngineValueFactory valueFactory = new EngineValueFactory(favorBigDecimal);
+    ValueSupportFactory valueFactory = new ValueSupportFactory(favorBigDecimal);
 
     // Execute interpretation securely
     Optional<JoshProgram> programResult = executeInterpretation(
@@ -211,7 +211,7 @@ public class JoshSimWorkerHandler implements HttpHandler {
     return OutputStepsParser.parseForWasmOrRemote(outputSteps);
   }
 
-  private Optional<JoshProgram> executeInterpretation(EngineValueFactory valueFactory,
+  private Optional<JoshProgram> executeInterpretation(ValueSupportFactory valueFactory,
         EngineGeometryFactory geometryFactory, ParseResult parsed,
         InputOutputLayer inputOutputLayer, HttpServerExchange httpServerExchange, String apiKey) {
     try {
@@ -242,7 +242,7 @@ public class JoshSimWorkerHandler implements HttpHandler {
    *     should be exported.
    * @return true on success, false on failure (400 response set).
    */
-  private boolean executeSimulation(EngineValueFactory valueFactory,
+  private boolean executeSimulation(ValueSupportFactory valueFactory,
         EngineGeometryFactory geometryFactory, String externalData, JoshProgram program,
         String simulationName, HttpServerExchange httpServerExchange, String apiKey,
         Optional<Set<Integer>> outputSteps) {
--- a/src/main/java/org/joshsim/cloud/JoshSimWorkerHandler.java
+++ b/src/main/java/org/joshsim/cloud/JoshSimWorkerHandler.java
@@ -16,12 +16,10 @@ import java.io.IOException;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
-import org.apache.sis.referencing.CRS;
 import org.joshsim.JoshSimFacadeUtil;
 import org.joshsim.engine.geometry.EngineGeometryFactory;
 import org.joshsim.engine.geometry.grid.GridGeometryFactory;
 import org.joshsim.engine.value.engine.ValueSupportFactory;
-import org.joshsim.geo.geometry.EarthGeometryFactory;
 import org.joshsim.lang.interpret.JoshProgram;
 import org.joshsim.lang.io.InputOutputLayer;
 import org.joshsim.lang.io.SandboxExportCallback;
@@ -61,15 +59,7 @@ public class JoshSimWorkerHandler implements HttpHandler {
       throw new RuntimeException("Only sandboxed mode is supported at this time.");
     }
 
-    if (crs.isPresent()) {
-      try {
-        geometryFactory = new EarthGeometryFactory(CRS.forCode(crs.get()));
-      } catch (Exception e) {
-        throw new RuntimeException("Failed to parse CRS: " + e);
-      }
-    } else {
-      geometryFactory = new GridGeometryFactory();
-    }
+    geometryFactory = new GridGeometryFactory();
   }
 
   /**
--- a/src/main/java/org/joshsim/command/RunCommand.java
+++ b/src/main/java/org/joshsim/command/RunCommand.java
@@ -3,8 +3,8 @@
  * Command line interface handler for running Josh simulations.
  *
  * <p>This class implements the 'run' command which executes a specified simulation from a Josh
- * script file. It supports both grid-based and earth-based coordinate reference systems, and can
- * process patches either serially or in parallel.</p>
+ * script file. It processes patches either serially or in parallel using grid-based coordinate
+ * space.</p>
  *
  * @license BSD-3-Clause
  */
@@ -20,7 +20,6 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.Callable;
-import org.apache.sis.referencing.CRS;
 import org.joshsim.JoshSimCommander;
 import org.joshsim.JoshSimFacadeUtil;
 import org.joshsim.compat.CompatibilityLayerKeeper;
@@ -33,7 +32,6 @@ import org.joshsim.engine.geometry.grid.GridGeometryFactory;
 import org.joshsim.engine.value.converter.Units;
 import org.joshsim.engine.value.engine.ValueSupportFactory;
 import org.joshsim.engine.value.type.EngineValue;
-import org.joshsim.geo.geometry.EarthGeometryFactory;
 import org.joshsim.lang.bridge.GridInfoExtractor;
 import org.joshsim.lang.bridge.ShadowingEntity;
 import org.joshsim.lang.interpret.JoshProgram;
@@ -58,8 +56,6 @@ import org.joshsim.util.ProgressUpdate;
 import org.joshsim.util.SharedRandom;
 import org.joshsim.util.SimulationMetadata;
 import org.joshsim.util.SimulationMetadataExtractor;
-import org.opengis.referencing.crs.CoordinateReferenceSystem;
-import org.opengis.util.FactoryException;
 import picocli.CommandLine.Command;
 import picocli.CommandLine.Mixin;
 import picocli.CommandLine.Option;
@@ -100,10 +96,6 @@ public class RunCommand implements Callable<Integer> {
 
   @Parameters(index = "1", description = "Simulation to run")
   private String simulation;
-
-  @Option(names = "--crs", description = "Coordinate Reference System", defaultValue = "")
-  private String crs;
-
 
   @Option(names = "--replicates", description = "Number of replicates to run", defaultValue = "1")
   private int replicates = 1;
@@ -256,18 +248,7 @@ public class RunCommand implements Callable<Integer> {
 
     // Parse output steps early for fail-fast validation
     final Optional<Set<Integer>> parsedOutputSteps = parseOutputSteps();
-    EngineGeometryFactory geometryFactory;
-    if (crs.isEmpty()) {
-      geometryFactory = new GridGeometryFactory();
-    } else {
-      CoordinateReferenceSystem crsRealized;
-      try {
-        crsRealized = CRS.forCode(crs);
-      } catch (FactoryException e) {
-        throw new RuntimeException(e);
-      }
-      geometryFactory = new EarthGeometryFactory(crsRealized);
-    }
+    EngineGeometryFactory geometryFactory = new GridGeometryFactory();
 
     // Parse custom parameters from command line
     Map<String, String> customParameters = parseCustomParameters();
--- a/src/main/java/org/joshsim/command/RunCommand.java
+++ b/src/main/java/org/joshsim/command/RunCommand.java
@@ -31,12 +31,15 @@ import org.joshsim.engine.geometry.ExtentsUtil;
 import org.joshsim.engine.geometry.PatchBuilderExtentsBuilder;
 import org.joshsim.engine.geometry.grid.GridGeometryFactory;
 import org.joshsim.engine.value.converter.Units;
-import org.joshsim.engine.value.engine.EngineValueFactory;
+import org.joshsim.engine.value.engine.ValueSupportFactory;
 import org.joshsim.engine.value.type.EngineValue;
 import org.joshsim.geo.geometry.EarthGeometryFactory;
 import org.joshsim.lang.bridge.GridInfoExtractor;
 import org.joshsim.lang.bridge.ShadowingEntity;
 import org.joshsim.lang.interpret.JoshProgram;
+import org.joshsim.lang.interpret.RecursiveValueResolverFactory;
+import org.joshsim.lang.interpret.TimedRecursiveValueResolverFactory;
+import org.joshsim.lang.interpret.ValueResolverFactory;
 import org.joshsim.lang.io.InputGetterStrategy;
 import org.joshsim.lang.io.InputOutputLayer;
 import org.joshsim.lang.io.JvmInputOutputLayerBuilder;
@@ -77,6 +80,20 @@ import picocli.CommandLine.Parameters;
 public class RunCommand implements Callable<Integer> {
   private static final int MINIO_ERROR_CODE = 100;
   private static final int UNKNOWN_ERROR_CODE = 404;
+
+  /**
+   * Builds the appropriate ValueResolverFactory based on whether profiling is enabled.
+   *
+   * @param enableProfiler True to use timed resolution for evalDuration support, false otherwise.
+   * @return A ValueResolverFactory configured for the requested profiling mode.
+   */
+  private static ValueResolverFactory buildValueResolverFactory(boolean enableProfiler) {
+    if (enableProfiler) {
+      return new TimedRecursiveValueResolverFactory();
+    } else {
+      return new RecursiveValueResolverFactory();
+    }
+  }
 
   @Parameters(index = "0", description = "Path to file to validate")
   private File file;
@@ -166,6 +183,13 @@ public class RunCommand implements Callable<Integer> {
                   + "operations will use this seed to produce deterministic results."
   )
   private Long seed = null;
+
+  @Option(
+      names = "--enable-profiler",
+      description = "Enable evalDuration profiling to capture attribute resolution timing.",
+      defaultValue = "false"
+  )
+  private boolean enableProfiler;
 
   /**
    * Parses custom parameter command-line options.
@@ -332,9 +356,12 @@ public class RunCommand implements Callable<Integer> {
     compatLayer.setExportQueueCapacity(exportQueueSize);
     CompatibilityLayerKeeper.set(compatLayer);
 
-    // Set up EngineValueFactory
+    // Set up ValueSupportFactory
     boolean favorBigDecimal = !useFloat64;
-    EngineValueFactory valueFactory = new EngineValueFactory(favorBigDecimal);
+    ValueSupportFactory valueFactory = new ValueSupportFactory(
+        favorBigDecimal,
+        buildValueResolverFactory(enableProfiler)
+    );
 
     // Extract grid information for Earth-space detection (similar to JoshSimFacade)
     MutableEntity simEntityRaw = program.getSimulations().getProtoype(simulation).build();
--- a/src/main/java/org/joshsim/command/RunCommand.java
+++ b/src/main/java/org/joshsim/command/RunCommand.java
@@ -248,7 +248,7 @@ public class RunCommand implements Callable<Integer> {
 
     // Parse output steps early for fail-fast validation
     final Optional<Set<Integer>> parsedOutputSteps = parseOutputSteps();
-    EngineGeometryFactory geometryFactory = new GridGeometryFactory();
+    final EngineGeometryFactory geometryFactory = new GridGeometryFactory();
 
     // Parse custom parameters from command line
     Map<String, String> customParameters = parseCustomParameters();
--- a/src/main/java/org/joshsim/engine/geometry/grid/GridCrsDefinition.java
+++ b/src/main/java/org/joshsim/engine/geometry/grid/GridCrsDefinition.java
@@ -11,11 +11,11 @@ import org.joshsim.engine.geometry.PatchBuilderExtents;
  *
  * <p>This class carries two cell size values to handle the count-conversion path:
  * <ul>
- *   <li>{@link #getIndexCellSize()} — the cell width in index/coordinate space.
- *       After count-conversion (WGS84 degrees → count coordinates), this is {@code 1}.
- *       On the Earth-space path (no count-conversion), this equals the meter value.</li>
- *   <li>{@link #getOriginalCellSizeMeters()} — the user-specified cell size in meters,
- *       preserved across count-conversion. Always physically meaningful.</li>
+ *   <li>{@link #getCellSizeCoords()} — the cell width in the grid's coordinate space.
+ *       After count-conversion (WGS84 degrees → count indices), this is {@code 1}.
+ *       For projected or meter-space grids, this equals {@link #getCellSizeMeters()}.</li>
+ *   <li>{@link #getCellSizeMeters()} — the physical cell size in meters, always the
+ *       user-specified value regardless of coordinate conversions.</li>
  * </ul>
  * </p>
  */
@@ -23,7 +23,7 @@ public class GridCrsDefinition {
   private final String name;
   private final String baseCrsCode;
   private final PatchBuilderExtents extents;
-  private final BigDecimal indexCellSize;
+  private final BigDecimal coordCellSize;
   private final BigDecimal cellSizeMeters;
 
   /**
@@ -35,7 +35,7 @@ public class GridCrsDefinition {
    * @param name The name of the grid system
    * @param baseCrsCode The EPSG code or identifier for base CRS
    * @param extents The grid extents in base CRS coordinates
-   * @param cellSize Cell size in both index space and meters (they are equal)
+   * @param cellSize Cell size in both coordinate space and meters (they are equal)
    */
   public GridCrsDefinition(
       String name,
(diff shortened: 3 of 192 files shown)


src/main/java/org/joshsim/command/RunCommand.java
```diff
@@ -256,18 +248,7 @@ public Integer call() {
...
-    } else {
-      CoordinateReferenceSystem crsRealized;
-      try {
-        crsRealized = CRS.forCode(crs);
-      } catch (FactoryException e) {
-        throw new RuntimeException(e);
-      }
-      geometryFactory = new EarthGeometryFactory(crsRealized);
-    }
+    final EngineGeometryFactory geometryFactory = new GridGeometryFactory();
```
GondekNP, line 251: EngineGeometryFactory now just always lives in GridSpace only, no longer needs CRS information

src/main/java/org/joshsim/engine/geometry/grid/GridCrsDefinition.java
```diff
@@ -9,51 +9,105 @@
...
+      PatchBuilderExtents extents,
+      BigDecimal indexCellSize,
+      BigDecimal cellSizeMeters) {
 
     this.name = Objects.requireNonNull(name, "Name cannot be null");
     this.baseCrsCode = Objects.requireNonNull(baseCrsCode, "Base CRS code cannot be null");
     this.extents = Objects.requireNonNull(extents, "Extents cannot be null");
-    this.cellSize = Objects.requireNonNull(cellSize, "Cell size cannot be null");
-    this.cellSizeUnits = Objects.requireNonNull(cellSizeUnits, "Cell size units cannot be null");
+    this.indexCellSize = Objects.requireNonNull(indexCellSize, "Index cell size cannot be null");
```
GondekNP: Arguably we could do away with this, right, because if we stay in 'Grid Space` forever there will never be a situation where our indexCellSize =/= 0?

src/main/java/org/joshsim/engine/geometry/grid/GridCrsDefinition.java
```diff
@@ -9,51 +9,105 @@
...
+   * Gets the original cell size in meters before any count-space conversion.
+   *
+   * <p>This is always the user-specified physical cell size. When no count-conversion
+   * occurred, this equals {@link #getIndexCellSize()}. When degree extents were converted
+   * to count coordinates, {@code getIndexCellSize()} returns {@code 1} while this method
+   * returns the original meter value (e.g., 30m).</p>
+   *
+   * @return the cell size in meters
+   */
+  public BigDecimal getOriginalCellSizeMeters() {
```
GondekNP: If we get rid of `indexCellSize` we could just change to `getCellSizeMeters()`

src/main/java/org/joshsim/engine/simulation/PatchSpatialIndex.java
```diff
@@ -67,6 +70,7 @@ class PatchSpatialIndex {
       this.maxX = BigDecimal.ZERO;
       this.maxY = BigDecimal.ZERO;
       this.cellSize = BigDecimal.ONE;
+      this.cellSizeInMeters = cellSizeMeters != null ? cellSizeMeters : BigDecimal.ONE;
```
GondekNP: Seems like a dangerous fallback?

src/main/java/org/joshsim/engine/simulation/PatchSpatialIndex.java
```diff
@@ -111,28 +124,51 @@ private GridParameters analyzePatches(Map<GeoKey, Entity> patches) {
...
+        foundMaxX = centerX;
       }
       if (foundMaxY == null || centerY.compareTo(foundMaxY) > 0) {
         foundMaxY = centerY;
       }
 
-      if (foundCellSize == null && geom.getOnGrid() != null) {
-        foundCellSize = geom.getOnGrid().getWidth();
+      if (metersCellSize == null && geom.getOnGrid() != null) {
+        metersCellSize = geom.getOnGrid().getWidth();
```
GondekNP: Again, seems like a dangerous fallback

src/main/java/org/joshsim/engine/simulation/PatchSpatialIndex.java
```diff
@@ -111,28 +124,51 @@ private GridParameters analyzePatches(Map<GeoKey, Entity> patches) {
...
       return null;
     }
 
-    return new GridParameters(foundMinX, foundMinY, foundMaxX, foundMaxY, foundCellSize);
+    boolean hasHorizSpacing = secondMinX != null && secondMinX.compareTo(foundMinX) > 0;
+    boolean hasVertSpacing = secondMinY != null && secondMinY.compareTo(foundMinY) > 0;
+
+    BigDecimal coordCellSize;
+    if (hasHorizSpacing) {
+      coordCellSize = secondMinX.subtract(foundMinX);
```
GondekNP: In grid space the x and y spacing should be the same - this check for vertical and horizontal spacing is probably overkill but there might be cases where we linear grids?

src/main/java/org/joshsim/engine/simulation/PatchSpatialIndex.java
```diff
@@ -111,28 +124,51 @@ private GridParameters analyzePatches(Map<GeoKey, Entity> patches) {
...
+    } else if (hasVertSpacing) {
+      coordCellSize = secondMinY.subtract(foundMinY);
+    } else {
+      coordCellSize = metersCellSize;
```
GondekNP: regarding the dangerous fallback comment at R150 - I suppose we need something here to account for the case where we have only one grid cell / patch. But I suppose maybe it's okay if we can't derive a grid spacing, and upstream hasn't populated one, but then at Line 154 we should probably just throw?

src/main/java/org/joshsim/lang/interpret/machine/SingleThreadEventHandlerMachine.java
```diff
@@ -516,7 +516,14 @@ public EventHandlerMachine createEntity(String entityType) {
 
   @Override
   public EventHandlerMachine executeSpatialQuery(ValueResolver resolver) {
-    EngineValue distance = convert(pop(), METER_UNITS);
+    EngineValue rawDistance = pop();
+    if (rawDistance.getUnits().equals(Units.EMPTY)) {
+      throw new IllegalArgumentException(
+          "Spatial query 'within' requires a distance with explicit units (e.g., '60 m'), "
```
GondekNP, line 522: We force users to define a `within` unit now, to avoid confusion with reaching over by number of grid cells for example - @lucialayr and I thought this was an okay tradeoff because one could always just calculate the number of grid cell units dynamically in terms of meters and drop into the `within` clause

src/test/java/org/joshsim/conformance/JoshConformanceTest.java
```diff
@@ -93,11 +95,16 @@ static Stream<TestInfo> discoverCriticalTests() throws Exception {
...
   private void runJoshTest(TestInfo test) throws Exception {
-    ProcessBuilder pb = new ProcessBuilder(
+    List<String> args = new ArrayList<>(List.of(
         "java", "-jar", JOSH_JAR,
         "run", test.path.toString(), test.simulationName,
         "--seed", "42"
-    );
+    ));
+    if (test.metadata.crs != null) {
+      args.add("--crs");
```
GondekNP: Needs to be updated, after all we are no longer accepting --crs at runtime

GondekNP (commented):
Some comments for claude!

GondekNP:
(Claude Plan - After Review)

## Plan to address review comments

### Change 1: Rename `GridCrsDefinition` getters for clarity

After discussion: both fields are needed (they differ on the count-conversion path), but the names are misleading. The two fields represent different concepts:

| Path | `indexCellSize` (current) | `cellSizeMeters` (current) | Equal? |
|------|--------------------------|---------------------------|--------|
| Count-space (WGS84 prod) | `1` (count units) | `30` (meters) | No |
| Earth-space (preprocessing) | `30` (meters) | `30` (meters) | Yes |
| Projected/pure meter | `10` (meters) | `10` (meters) | Yes |

- "Index" is misleading — on projected grids, the value is meters, not indices
- "Original" prefix is misleading — nothing was overwritten on non-count-conversion paths

**Renames:**
- `getIndexCellSize()` → `getCellSizeCoords()` — "cell width in the grid's coordinate space"
- `getOriginalCellSizeMeters()` → `getCellSizeMeters()` — "physical cell size in meters"
- Internal field: `indexCellSize` → `coordCellSize`

### Change 2: Tighten `PatchSpatialIndex` fallbacks

After the bug fix, `cellSizeMeters` is always passed from `TimeStep`. The fallbacks are dead code that would mask bugs.

- Add `Objects.requireNonNull(cellSizeMeters)` at top of constructor
- Remove `cellSizeMeters != null ? cellSizeMeters : BigDecimal.ONE` ternaries (now guaranteed non-null)
- Remove `getOnGrid().getWidth()` fallback extraction (line 150) — meter size comes from constructor, not patch geometry
- Change `return null` at line 154 to `throw IllegalStateException` as safety net

### Change 3: Remove `--crs` from conformance test runner

`RunCommand` no longer accepts `--crs`, and no test files use `# @crs:` metadata. Remove:
- `if (test.metadata.crs != null)` block from `JoshConformanceTest.runJoshTest()`
- `crs` field from `TestMetadata` (field, constructor param, parsing case)
- Any `crs` references in `TestMetadataTest`

### Change 4: No action (informational comments)

- RunCommand acknowledgment ✓
- SingleThreadEventHandlerMachine unit enforcement ✓  
- PatchSpatialIndex horiz/vert spacing — keeps handling 1×N grids, worth the 3 lines

src/main/java/org/joshsim/engine/simulation/PatchSpatialIndex.java
```diff
@@ -271,7 +307,7 @@ private List<Entity> queryCandidatesForCircle(GridShape circle) {
     int centerGridX = worldToGridX(centerX);
     int centerGridY = worldToGridY(centerY);
 
-    double radiusInGridCells = diameter.doubleValue() / (2.0 * cellSize.doubleValue());
+    double radiusInGridCells = diameter.doubleValue() / (2.0 * cellSizeInMeters.doubleValue());
```
GondekNP: this is the reason for threading GridCrsDefinition all the way through

GondekNP:
(Codex 5.3 Review - Adversarial Review through OpenCode)

# Spatial Query Fix Review (PR #370)

Branch reviewed: `feat/simplify_grid_route`

Base compared: `origin/dev`

PR: https://github.com/SchmidtDSE/josh/pull/370

Date: 2026-04-10

## Review scope and method

This review covers runtime geometry routing, spatial query radius conversion, CRS/cell-size threading, API contract changes, and test coverage added in this branch.

Reviewed inputs include:

- Full branch diff versus `origin/dev` (42 files changed)
- Main production files in `command`, `cloud`, `engine`, `lang`, `geo`, and `precompute`
- Added/updated unit tests and conformance tests
- PR metadata and CI checks from GitHub
- Local validation runs: `./gradlew test`, `./gradlew checkstyleTest`, `./gradlew conformanceTest`

## Executive assessment

Verdict: **Approved with minor follow-ups**.

The branch successfully does two important things:

1. Removes the runtime CRS/Earth geometry branch from `run` path, simplifying execution and reducing dead branching.
2. Preserves and correctly threads physical meter cell size (`cellSizeMeters`) into spatial `within` queries, fixing the radius conversion bug that previously over-expanded neighborhoods.

No blocker-level regressions were found in reviewed code paths. The largest remaining issues are consistency/documentation cleanup and one edge-case fallback that is conceptually fragile but low risk in common usage.

## What changed (confirmed)

- Runtime now always uses `GridGeometryFactory`:
  - `RunCommand` no longer exposes `--crs`.
  - `JoshSimWorkerHandler` no longer uses CRS to switch runtime geometry factory.
- Engine/geo decoupling improved:
  - `EngineGeometry` no longer requires `getOnEarth()`.
  - Reverse dependency from `engine` core to `geo` Earth shape contracts removed.
- `GridCrsDefinition` split into explicit values:
  - `getCellSizeCoords()` for coordinate-space spacing.
  - `getCellSizeMeters()` for physical spacing.
- `GridFromSimFactory` now preserves original meter size before count-space overwrite.
- `PatchSpatialIndex` now:
  - Requires non-null `cellSizeMeters`.
  - Uses coordinate spacing inferred from centers for indexing.
  - Uses `cellSizeMeters` for meter-to-grid radius conversion.
  - Removes the old geometry-width fallback that caused incorrect conversions.
- Spatial query contract hardened:
  - Unitless `within` distance now throws a clear actionable error.
- Conformance harness simplified:
  - `--crs` removed from conformance run path.
  - `@crs` metadata parsing removed.
- New conformance coverage added for meter and WGS84 scenarios (11 new files).

## Findings by severity

### 1) Stale runtime CRS comments and dead optional parameter

- Severity: **Minor**
- Status: **Introduced by this branch**
- Affected files:
  - `src/main/java/org/joshsim/command/RunCommand.java`
  - `src/main/java/org/joshsim/cloud/JoshSimWorkerHandler.java`
  - `src/main/java/org/joshsim/cloud/JoshParseHandler.java`
- Details:
  - Runtime behavior now ignores CRS for `run`, but docs/comments still mention CRS in places.
  - `JoshSimWorkerHandler` constructor still accepts `Optional<String> crs` although runtime factory selection no longer uses it.
  - `JoshParseHandler` still carries a redundant `if (crs.isPresent())` branch that assigns the same `GridGeometryFactory` either way.
- Risk:
  - Reader confusion, future maintenance mistakes, and API drift.
- Recommendation:
  - Remove unused constructor arg and dead branch, then update Javadocs/help text to match actual runtime behavior.

### 2) Degenerate-grid fallback mixes coordinate and meter semantics

- Severity: **Minor (watch item)**
- Status: **Introduced/reshaped in this branch**
- Affected file:
  - `src/main/java/org/joshsim/engine/simulation/PatchSpatialIndex.java`
- Details:
  - In analysis fallback paths where coordinate spacing cannot be inferred (for example degenerate shapes), inferred coordinate `cellSize` may default to `cellSizeMeters`.
  - This can conceptually mix spaces if coordinate system is count-space (`coordCellSize = 1`) while meters are not 1.
- Risk:
  - Low in normal rectangular grids; potentially confusing/incorrect in edge geometries.
- Recommendation:
  - Prefer using `GridCrsDefinition.getCellSizeCoords()` as fallback for coordinate spacing, reserving `cellSizeMeters` only for distance conversion.

### 3) Local-vs-CI conformance mismatch

- Severity: **Informational**
- Status: **Pre-existing test/environment discrepancy**
- Details:
  - PR checks in GitHub are green.
  - Local `./gradlew conformanceTest` reported multiple failures (including known conformance and performance tests), while newly added spatial tests passed locally.
- Risk:
  - Medium process risk: reproducibility friction, but no clear branch regression signal.
- Recommendation:
  - Capture/compare CI runtime inputs (data files, env, flags, machine profile) to reduce ambiguity in local failure interpretation.

## Regression and compatibility assessment

- Runtime behavior change is intentional and consistent with branch goal:
  - `run` no longer accepts CRS path switching.
  - `preprocess` CRS behavior remains in place.
- Breaking user-facing changes (intentional):
  - Unitless spatial radius in `within` now errors; callers must specify units (`m`).
  - `RunCommand --crs` removed.
- API changes (intentional):
  - `GridCrsDefinition` getter renames/removals align naming with actual semantics.

No unintended runtime regressions were identified in reviewed code paths.

## Test coverage assessment

Coverage added in this branch is strong for the fixed defect:

- WGS84 conformance matrix validates latitude and resolution independence.
- Meter-grid tests validate expected geometric neighbor counts.
- Unit tests updated across `GridCrsDefinition`, `Replicate`, `TimeStep`, and geo mapping paths.

Net: the new tests are appropriately targeted and materially improve confidence in radius conversion correctness.

## Recommended follow-up patch set

1. **Code/doc cleanup (minor)**
   - Remove unused runtime CRS constructor param and dead branching.
   - Update stale comments/Javadocs/help text referencing runtime CRS behavior.

2. **Edge-case fallback cleanup (minor)**
   - In `PatchSpatialIndex`, ensure coordinate spacing fallback uses coordinate-space size source.

3. **Reproducibility note (informational)**
   - Document expected local prerequisites for stable `conformanceTest` parity with CI.

## Final decision

The branch is technically sound and addresses the core bug and simplification goals effectively. I recommend merge after (or with) minor cleanup follow-ups noted above.

GondekNP:
@sampottinger - going to take another look tomorrow AM with some fresh eyes, but ready for your review!

src/main/java/org/joshsim/command/RunCommand.java
sampottinger: I like how much cleaner this made things…

src/main/java/org/joshsim/engine/geometry/grid/GridCrsDefinition.java
```diff
@@ -9,51 +9,100 @@
  * Definition of a grid coordinate reference system.
  * Holds parameters needed to map between grid coordinates and a real-world CRS.
  *
- * <p>The cell size is always specified in meters, regardless of the base CRS units.
- * When using a geographic CRS (such as EPSG:4326), the cell size will be applied
- * in the intermediate projected space.</p>
+ * <p>This class carries two cell size values to handle the count-conversion path:
+ * <ul>
+ *   <li>{@link #getCellSizeCoords()} — the cell width in the grid's coordinate space.
+ *       After count-conversion (WGS84 degrees → count indices), this is {@code 1}.
```
sampottinger: There might be some issues with the — and → for screen readers and some IDEs. We might want to remove.

src/main/java/org/joshsim/cloud/JoshSimWorkerHandler.java
```diff
@@ -61,15 +59,7 @@ public JoshSimWorkerHandler(CloudApiDataLayer apiInternalLayer, boolean sandboxe
...
-    if (crs.isPresent()) {
-      try {
-        geometryFactory = new EarthGeometryFactory(CRS.forCode(crs.get()));
-      } catch (Exception e) {
-        throw new RuntimeException("Failed to parse CRS: " + e);
-      }
-    } else {
-      geometryFactory = new GridGeometryFactory();
-    }
+    geometryFactory = new GridGeometryFactory();
```
sampottinger, line 62: Like how clean this is now.

src/main/java/org/joshsim/engine/geometry/grid/GridCrsDefinition.java
```diff
@@ -9,51 +9,100 @@
...
-  public BigDecimal getCellSize() {
-    return cellSize;
+  /**
+   * Gets the cell width in the grid's coordinate space.
+   *
+   * <p>After count-conversion (WGS84 degrees → count indices), this returns {@code 1}.
+   * For projected or meter-space grids, this equals {@link #getCellSizeMeters()}.
+   * Use this for grid layout, coordinate transforms, and patch construction.</p>
+   *
+   * @return cell width in coordinate space
```
sampottinger: Coordinate space … I wonder if we should grid space or cell space? WDYT?

src/main/java/org/joshsim/engine/geometry/grid/GridCrsDefinition.java
```diff
@@ -93,22 +142,22 @@ public BigDecimal[] gridToCrsCoordinates(BigDecimal gridX, BigDecimal gridY) {
    */
   public BigDecimal[] crsToGridCoordinates(BigDecimal crsX, BigDecimal crsY) {
     BigDecimal gridX =
```
sampottinger: Niggle but I wonder what you think of this…

```
BigDecimal deoriginedX = crsX.subtract(extents.getTopLeftX());
BigDecimal gridX = deoriginedX.divide(
    coordCellSize,
    10,
    RoundingMode.HALF_UP
);
```

Sorry I’m missing something obvious but I can’t recall what 10 means here. :P

Could also extract to a private method doing this for a single coordinate to reduce code duplication but I see opinions either way.

src/main/java/org/joshsim/engine/geometry/grid/GridCrsDefinition.java
```diff
@@ -93,22 +142,22 @@ public BigDecimal[] gridToCrsCoordinates(BigDecimal gridX, BigDecimal gridY) {
...
-        crsY.subtract(extents.getTopLeftY()).divide(cellSize, 10, RoundingMode.HALF_UP);
+        crsY.subtract(extents.getTopLeftY()).divide(coordCellSize, 10, RoundingMode.HALF_UP);
     return new BigDecimal[] {gridX, gridY};
   }
 
   @Override
   public String toString() {
     return String.format(
-        "GridCrsDefinition[name=%s, extents=(%s,%s to %s,%s), cellSize=%s %s]",
+        "GridCrsDefinition[name=%s, extents=(%s,%s to %s,%s), coordCellSize=%s, cellSizeMeters=%s]",
```
sampottinger: I wonder if we should hoist this up to a static final String just since it’s getting a little longer. What do you think?

src/main/java/org/joshsim/engine/simulation/PatchSpatialIndex.java
```diff
@@ -92,13 +100,21 @@ class PatchSpatialIndex {
...
    * Analyzes patches to determine grid parameters.
+   *
+   * <p>Infers coordinate spacing from the smallest gap between distinct center
+   * coordinates in either axis.</p>
+   *
+   * @param patches the patches to analyze
+   * @param cellSizeMeters cell size in meters (non-null)
    */
-  private GridParameters analyzePatches(Map<GeoKey, Entity> patches) {
+  private GridParameters analyzePatches(Map<GeoKey, Entity> patches, BigDecimal cellSizeMeters) {
```
sampottinger: I think this makes sense. Just looking back at it and feeling bad, I wonder if I maybe should have called this `inferGridParametersByCenterpoint` instead? Just reading the code above I wasn’t quite sure.

src/main/java/org/joshsim/engine/simulation/PatchSpatialIndex.java
```diff
@@ -111,28 +127,48 @@ private GridParameters analyzePatches(Map<GeoKey, Entity> patches) {
       BigDecimal centerY = geom.getCenterY();
 
       if (foundMinX == null || centerX.compareTo(foundMinX) < 0) {
+        secondMinX = foundMinX;
         foundMinX = centerX;
+      } else if (secondMinX == null || (centerX.compareTo(foundMinX) > 0
```
sampottinger: I wonder if these might be slightly more readable if we unpack it a bit…

```
/**
 * Class to find the minimum and second smallest to that minimum value iteratively.
 */
public class MinKeepers {

  private BigDecimal min;
  private BigDecimal secondToMin;

  public MinKeeper() {
    min = null;
    secondToMin = null;
  }

  public void evaluate(BigDecimal candidate) {
    boolean noMin = min == null;
    if (noMin) {
      acceptNewMin(candidate);
      return;
    }

    boolean lessThanMin = centerX.compareTo(foundMinX) < 0;
    if (lessThanMin) {
      acceptNewMin(candidate);
      return;
    }

    // …
  }

  public BigDecimal getMin();

  public BigDecimal getSecondToMin();

  private void acceptNewMin(BigDecimal candidate)

  private void acceptNewSecondToMin(BigDecimal candidate)

}
```

Just checking I don’t think we are concerned where there is only one value processed right? I also apologize it’s probably obvious but why do we need both min and second min?

src/main/java/org/joshsim/engine/simulation/TimeStep.java
```diff
@@ -48,13 +66,30 @@ private PatchSpatialIndex getSpatialIndex() {
...
 
+  /**
+   * Gets the grid CRS definition, if available.
+   *
+   * <p>Available when the replicate was constructed from a PatchSet (the standard path).
+   * Null when constructed from raw patch maps (test/fallback path).</p>
+   *
+   * @return the grid CRS definition, or null
+   */
+  public GridCrsDefinition getGridCrsDefinition() {
```
sampottinger: I know there is some disagreements about it in the Java community but I wonder if you think `Optional` is more aligned with the rest of the codebase?

src/main/java/org/joshsim/lang/interpret/machine/SingleThreadEventHandlerMachine.java
sampottinger: Could one use within 5 count or would that be a prohibited too?

src/test/java/org/joshsim/engine/geometry/grid/GridCrsDefinitionTest.java
```diff
@@ -193,11 +175,11 @@ void crsToGridAndBackShouldBeIdentity() {
   @Test
   void toStringFormatIsCorrect() {
     String expected = String.format(
-        "GridCrsDefinition[name=%s, extents=(%s,%s to %s,%s), cellSize=%s %s]",
+        "GridCrsDefinition[name=%s, extents=(%s,%s to %s,%s), coordCellSize=%s, cellSizeMeters=%s]",
```
sampottinger: Getting a bit longer, we might consider hoisting to a final static… could also reference from the class being tested but I see opinions either way.

sampottinger (changes requested):
Some minor things…

src/main/java/org/joshsim/engine/simulation/PatchSpatialIndex.java
sampottinger: Ah I see this is why we need both… Might consider adding a short comment.

GondekNP:
(Claude response to review)

**Re: sampottinger's question on `within 5 count`** (SingleThreadEventHandlerMachine.java:522)

`within 5 count` would pass the `Units.EMPTY` check (it has units), but then `convert(rawDistance, METER_UNITS)` would attempt count-to-meters conversion. Since no such conversion is registered, it would throw -- which is the correct behavior since "count" is not a distance unit.

The guard catches unitless values explicitly with an actionable error message, and the existing conversion pipeline rejects nonsensical unit conversions downstream.

---

**Changes pushed addressing sampottinger's review:**

1. **PatchSpatialIndex: pass grid cell size explicitly, remove spacing inference** (`eb1e5569`) — Now that `GridCrsDefinition` propagates both cell sizes through the full chain, the `secondMinX`/`secondMinY` logic that re-derived coordinate spacing from patch centers was redundant. `analyzePatches` simplified to `findPatchBounds` (min/max only). Addresses comments #5, #6, #7.

2. **Rename `getCellSizeCoords` → `getCellSizeGrid`** (`79260e79`) — Adopts "grid space" terminology. Also includes:
   - Replaced unicode chars (`→`, `—`) with ASCII in javadoc (comment #1)
   - Extracted `CRS_COORDINATE_SCALE = 10` constant + `crsToGridAxis()` helper (comment #3)
   - Hoisted `TO_STRING_FORMAT` to `static final` in both production and test (comments #4, #10)
   - Wrapped `gridCrsDefinition` in `Optional` in `TimeStep` (comment #8)

GondekNP:
(Claude from my IDE followup)

Follow-up: **Reject `count` units in spatial `within` queries** (`5a8c0238`)

Added explicit guard for `COUNT_UNITS` alongside the existing `EMPTY` check. Previously `within 5 count` would pass the unitless check but fail with an opaque conversion error downstream. Now throws immediately with an actionable message: `"Spatial query 'within' requires a physical distance unit (e.g., '60 m'), not 'count'"`.

GondekNP:
After some additional prompting locally:

1) Yes @sampottinger, `within 5 count` will now throw an `IllegalArgumentException`. I think this is for the best because I don't think most of those spatial queries will be resolution dependent (ie, I would want a user to express within 100m rather than within 2 count, which at 50m resolution is the same but when / if they change to 10m resolution is now suddenly 2/5 the reach). My assumption is that `within` should be used with biologically meaningful values, not grid conventions, but since we have this catch here we can always raise a warning or something like that if it seems preferred. 

@lucialayr and I came to the conclusion that a user could always define `const withinDistance = gridSize * 2` or something like that if they really wanted specifcially grid size dependent within distances, but overall it feels like an anitpattern to me (especially now that `joshpy` supports dynamically changing grid resolution with `GridSpec`). 

2) Also yes - I think referring to grid space as `Grid` matches the rest of the codebase, changed!

3) I realized that the whole business of inferring width defensively is no longer necessary, because we are no longer running simulations in Earth space anyways and the cell width can be just taken directly from GridCrsDefinition safely (it's deterministic at this point). Even better!

PR #371, opened 2026-04-11
Debug/integrated timeit

GondekNP:
#365 missed a proper integration test, and due to the fact that our actual tests were not computationally heavy, we (claude) didn't catch it before merge. 

The --enable-profiler flag's was wired into after program compilation, but instances are created during compilation. All resolvers were therefore non-timed and hardcoded to return 0ms.

Fix: Move ValueSupportFactory creation before getJoshProgram()  in RunCommand and add a JoshSimCommander.getJoshProgram() overload that accepts a caller-supplied factory, bypassing the hardcoded default in JoshSimFacade.interpret().

### DIFF
--- a/examples/features/eval_duration.josh
+++ b/examples/features/eval_duration.josh
@@ -36,6 +36,6 @@ end organism
 start organism Shrub
 
   cover.init = sample uniform from 0% to 5%
-  cover.step = prior.cover + sample uniform from 0% to 0.5%
+  cover.step = limit prior.cover + sample uniform from 0% to 0.5% to [0%, 100%]
 
 end organism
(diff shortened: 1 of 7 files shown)


GondekNP:
In the end - we switched to `nanoTime()` over `currentTimeMillis()` which as I understand has a bit more overhead, but since this is gated and opt-in seems okay. For the types of things we are testing it might end up being overkill, but at least in the toy examples it was hard to get a good signal from `currentTimeMillis()` (for eg, in the working example, we are instantiating a whole ton of agents which on their own init quickly but are death-by-a-thousand-cuts). the Ms approach might be just fine for 'real' profiling so flagging @sampottinger that perhaps we will want to revert that in particular!

Issue #374, opened 2026-04-11
Josh - remote target enhancement


GondekNP:
(Copy pasted issue document from slack canvas between @sampottinger and I - whacky formatting!)

**Challenge**

Cloud Run has hard ceilings (32 GiB memory, 60-min timeout, 8 vCPUs) that block HPC-scale simulations. This is problematic since most of our validation strategy is around capturing uncertainty through tons of replicates, and so we will need to run these models performantly and in parallel before long. While we can just ssh into a machine and handle it manually, this doesn't scale super well to the uncertainty capture case. 

**Proposal**

Two new execution targets for runRemote - Kubernetes and SSH - that let you submit fire-and-forget batch jobs to any K8s cluster or any machine with SSH and Java. Both are infrastructure-agnostic (no vendor lock-in), and both reuse the existing runRemote strategy pattern almost entirely. The user experience is just a new --target flag:

    joshsim runRemote simulation.josh MySim --target=nautilus --replicates=50
    joshsim runRemote simulation.josh MySim --target=lab-server --replicates=50

This covers our actual use cases without marrying us to GCP and keeps us interoperable (cc 2i2c Right to Replicate). K8s runs on GKE, EKS, AKS, Nautilus, self-hosted - whatever. SSH runs on literally anything. We considered GCE VM and Cloud Batch targets too but they're GCP-locked and K8s already covers GCP users via GKE. We can always add those later as convenience layers.
Important context: the current runRemote system is NOT Cloud Run-specific. It's a generic Undertow HTTP server. Anyone can run `java -jar joshsim-fat.jar server` on any machine and point runRemote at it. Cloud Run is just our hosted deployment - the code has zero coupling to it beyond a default URL. So the existing system is already portable, we're just adding fire-and-forget (submit and disconnect) and the ability to target bigger machines with platform-native scheduling.

**How parallelism works**

The existing LocalLeaderStrategy fans out like this: 1 replicate = 1 HTTP request = 1 Cloud Run worker. Workers are dumb - they run one replicate, don't even know their replicate number (always output as [0]), and the leader rewrites it. All coordination lives in the leader.
[...]

PR #375, opened 2026-04-11
Add download/list/delete operations to MinioHandler

GondekNP:
## Summary
- Extends `MinioHandler` with four new methods: `downloadFile()`, `downloadStream()`, `listObjects()`, `deleteObjects()`
- These are the read-side counterparts to the existing upload methods, needed for MinIO-based batch execution
- All methods respect the existing `basePath` convention via `constructObjectPath()`

## What changed

**`MinioHandler.java`** — 4 new public methods:
- `downloadFile(String objectPath, File destination)` — downloads object to local file via `getObject()` + `transferTo()`
- `downloadStream(String objectPath)` — returns `InputStream` for streaming reads (caller closes)
- `listObjects(String prefix)` — recursive listing via `listObjects()`, returns full object keys
- `deleteObjects(String prefix)` — lists then removes all objects under prefix, returns count

**`MinioHandlerTest.java`** — 7 new test cases:
- `downloadFile` writes correct contents to destination file
- `downloadFile` constructs correct object path with base path
- `downloadStream` returns readable InputStream
- `listObjects` returns matching keys
- `listObjects` returns empty list for no matches
- `deleteObjects` removes all listed objects
- `deleteObjects` returns zero and skips removal when nothing to delete

## Test plan
- [x] `./gradlew test` — all existing tests pass (no regressions)
- [x] `./gradlew checkstyleMain` + `checkstyleTest` — pass
- [x] `./gradlew fatJar` — builds cleanly
- [x] New unit tests for all 4 methods with mocked MinioClient

Part of #374 — PR 1 of 8 for enhanced remote execution targets.

### DIFF
--- a/src/main/java/org/joshsim/util/MinioHandler.java
+++ b/src/main/java/org/joshsim/util/MinioHandler.java
@@ -5,10 +5,17 @@
 package org.joshsim.util;
 
 import io.minio.BucketExistsArgs;
+import io.minio.GetObjectArgs;
+import io.minio.ListObjectsArgs;
 import io.minio.MakeBucketArgs;
 import io.minio.MinioClient;
+import io.minio.RemoveObjectArgs;
+import io.minio.Result;
 import io.minio.UploadObjectArgs;
+import io.minio.messages.Item;
 import java.io.File;
+import java.io.FileOutputStream;
+import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.List;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/joshsim/util/MinioHandler.java
```diff
@@ -141,6 +148,91 @@ private String constructObjectPath(String relativePath) {
...
+  }
+
+  /**
+   * Delete all objects under a given prefix.
+   *
+   * @param prefix The prefix to delete under (relative to base path)
+   * @return The number of objects deleted
+   * @throws Exception If deletion fails
+   */
+  public int deleteObjects(String prefix) throws Exception {
```
GondekNP, line 220: I think for safety sake, we don't really need deleteObjects - that's more of a lifecycle concern and out of scope for `josh`

GondekNP:
Upon some further thought - we already have --include-source, --include-jshd, etc, as ways to ensure the source is saved. We probably _do_ want to allow josh to delete intermediate data files (the copied over raw data or jshds in their temp location) so we will keep.

PR #381, opened 2026-04-12
Remove vestigial --upload-* flags

GondekNP:
## Summary
- Removed `--upload-source`, `--upload-config`, `--upload-data` flags and upload logic from `RunCommand`, `ValidateCommand`, and `RunRemoteCommand`
- Removed `saveToMinio()` static method from `JoshSimCommander` (no remaining callers)
- Removed `MinioOptions` mixin from `ValidateCommand` (only used for uploads)
- Deleted `RunCommandArtifactUploadTest` and `RunRemoteCommandArtifactUploadTest`
- Updated README and llms-full.txt to remove upload flag documentation

**Rationale:** `stageToMinio` now handles explicit uploads as a separate concern. joshpy bottling handles reproducibility. The `--upload-*` flags are vestigial.

**Breaking change:** Users relying on `--upload-source`, `--upload-config`, or `--upload-data` must switch to `stageToMinio`.

## Test plan
- [x] `./gradlew checkstyleMain` passes
- [x] `./gradlew test` passes (1813 tests)
- [x] Existing `runRemote` HTTP behavior unchanged


GondekNP (commented):
Much cleaner! We will probably need to make some `joshpy` tweaks down the line here, as I believe this was included in RunConfig based on available flags, but no live code was using it

PR #382, opened 2026-04-12
Add /runBatch endpoint for MinIO-staged execution

GondekNP:
## Summary
- New `JoshSimBatchHandler` (~230 lines) — Undertow handler for `POST /runBatch`
- Registered `/runBatch` path in `JoshSimServer`
- 11 unit tests for HTTP validation, API key, missing fields, error responses
- Documented all server endpoints in `llms-full.txt`
- Updated `K8S_PLANNING.md` with implementation details for PRs 3, 4, and 9

**How it works:** Inputs are pre-staged via `stageToMinio`. The handler downloads them to a temp dir, parses the Josh script, runs the simulation with parallel patches, and results land in MinIO via `minio://` export paths. Returns JSON status.

**Bug found & fixed:** Initial implementation had garbled CSV headers with parallel patches. Root cause: missing `CompatibilityLayerKeeper.set(new JvmCompatibilityLayer())` caused fallback to `EmulatedQueueService` (WASM shim) which processes export writes synchronously on the caller's thread instead of a dedicated writer thread.

**Integration tested:** `stageToMinio` → `POST /runBatch` → verified clean CSV in GCS bucket with correct data across all timesteps.

## Test plan
- [x] `./gradlew checkstyleMain checkstyleTest` passes
- [x] `./gradlew test` passes (1813 tests)
- [x] Integration: local server → stageToMinio → POST /runBatch → stageFromMinio → verify results
- [x] Existing `/runReplicate` and `/runReplicates` unchanged

### DIFF
--- a/src/main/java/org/joshsim/cloud/JoshSimBatchHandler.java
+++ b/src/main/java/org/joshsim/cloud/JoshSimBatchHandler.java
@@ -1,0 +1,217 @@
+/**
+ * Handler for batch simulation execution from pre-staged local files.
+ *
+ * <p>Implements the {@code /runBatch} endpoint. Unlike {@code /runReplicate} which streams code
+ * and data in the request body and returns results via wire format, this handler runs a simulation
+ * from a local directory of pre-staged files and results land in MinIO via {@code minio://} export
+ * paths configured in the Josh script. Staging is a separate concern handled by the caller (e.g.,
+ * {@code stageFromMinio} CLI command or K8s container entrypoint).</p>
+ *
+ * @license BSD-3-Clause
+ */
+
+package org.joshsim.cloud;
+
+import io.undertow.server.HttpHandler;
+import io.undertow.server.HttpServerExchange;
+import io.undertow.server.handlers.form.FormData;
+import io.undertow.server.handlers.form.FormDataParser;
+import io.undertow.server.handlers.form.FormParserFactory;
+import io.undertow.util.HttpString;
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.Map;
+import java.util.Optional;
+import org.joshsim.JoshSimFacadeUtil;
+import org.joshsim.compat.CompatibilityLayerKeeper;
+import org.joshsim.compat.JvmCompatibilityLayer;
+import org.joshsim.engine.geometry.grid.GridGeometryFactory;
+import org.joshsim.engine.value.engine.ValueSupportFactory;
+import org.joshsim.lang.interpret.JoshProgram;
+import org.joshsim.lang.io.InputOutputLayer;
+import org.joshsim.lang.io.JvmInputOutputLayerBuilder;
+import org.joshsim.lang.io.JvmMappedInputGetter;
+import org.joshsim.lang.parse.ParseResult;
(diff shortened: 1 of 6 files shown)


src/main/java/org/joshsim/cloud/JoshSimBatchHandler.java
```diff
@@ -0,0 +1,300 @@
...
+      sendJsonError(exchange, 500, jobId, "Batch execution failed: " + e.getMessage());
+    }
+
+    return Optional.of(apiKey);
+  }
+
+  private void executeBatchJob(String jobId, String simulation, String minioPrefix)
+      throws Exception {
+    // Ensure JVM compatibility layer is set so export queue services use real threads
+    // (without this, the fallback EmulatedCompatibilityLayer processes writes synchronously
```
GondekNP: Unneccessary exposition - not important to persist as a comment here

src/main/java/org/joshsim/cloud/JoshSimBatchHandler.java
```diff
@@ -0,0 +1,300 @@
...
+   * Constructs a new JoshSimBatchHandler.
+   *
+   * @param apiDataLayer The cloud API data layer for API key validation and logging.
+   */
+  public JoshSimBatchHandler(CloudApiDataLayer apiDataLayer) {
+    this.apiDataLayer = apiDataLayer;
+  }
+
+  @Override
+  public void handleRequest(HttpServerExchange exchange) throws Exception {
```
GondekNP, line 66: Seems like this should be a method on `HttpHandler` and not here? Should be DRY between runRemote and runBatch

src/main/java/org/joshsim/cloud/JoshSimBatchHandler.java
```diff
@@ -0,0 +1,300 @@
...
+          (step) -> { /* progress unused in batch mode */ },
+          false,
+          Optional.empty()
+      );
+    } finally {
+      deleteDirectory(tempDir);
+    }
+  }
+
+  private void stageFromMinio(MinioHandler minio, String prefix, File outputDir,
```
GondekNP: This violates our design direction. We want `runBatch` to be thin, basically JUST run the `simulations `from local (on the server side) and put the results wherever they need to go as defined in `josh`. 

At least in the k8s case (and from joshsim) we will handle the `stageFromMinio` call first to ensure the data are server-side-local, THEN run `runBatch`. Perhaps in the future that will feel like awkward ergonomics - but even in that case, we should definitely be calling a common utility with `stageFromMinio` rather than re-implementing here!

src/main/java/org/joshsim/cloud/JoshSimBatchHandler.java
```diff
@@ -0,0 +1,300 @@
...
+        throw new IOException("Failed to create directory: " + parentDir.getPath());
+      }
+
+      minio.downloadFile(key, destination);
+    }
+
+    output.printInfo("Staged " + keys.size() + " file(s) from MinIO prefix: " + prefix);
+  }
+
+  private Map<String, String> buildFileMapping(File directory) {
```
GondekNP: Evaluate if all of these methods (here and onward) should just be factored out common utilities between these server endpoints

PR #392, opened 2026-04-14
Remote profiler

### LINKED ISSUE #391: Adding remote profiler
# Remote profiler support
Expanding the testing for the profiling feature ("time it") that checks correctness both when profiling is enabled and when disabled before adding support for remote profiling.

## Background
We recently added support for the "time it" feature which is enabled on the jar through the `--enable-profiler` flag. This was introduced in PR #365 but expanded in #371 and #372. It is unclear if it is working fully right now, both locally and remote. Also, Josh Cloud currently runs without profiler enabled.

## Purpose
We should expand our CI / CD tests to check that time it works locally and remotely. We should then wire profiler support through the remote execution path so that self-hosted deployments can enable it per-request or server-wide, laying the groundwork for Josh Cloud to support it when deployed to main.

## Procedure
We will go through each component of this task using sub-agents: one component at a time and one subagent at a time. We should instruct AI to go through the following flow:

 - `component-planner` with path to the task markdown file and reminder not to modify code other than the task markdown file
 - `component-implementer` with path to the task markdown file and reminder not to worry about lint or style issues
 - `component-validator` with path to the task markdown file with instructions to update that file with indication if safe to continue
 - `component-concluder` with path to the task markdown file and instructions to make it clear if the component was successful
 - Pause if not safe to continue or the component was not successful, asking for feedback. Otherwise, continue to the next component.

Please sure there is a commit once per component with Claude as co-author. Please ask all agents to also review GitHub CI / CD actions before starting.

## Components
This task is decomposed into individual components which should be completed one at a time.

### 1. Test without profiler local
We should add to the "Test Additional Features" CI / CD checks. Specifically, we can use `examples/simulations/profiler_multi.josh` and we should run without `--enable-profiler` and we should ensure that the timings are reported but that it is all zero.

Details:
 - The simulation name is `ProfilerMultiExample`.
 - Run with `--replicates=1`.
 - The simulation writes to `/tmp/profiler_multi_josh.csv`. After running, immediately invoke the checker script (immediately failing the CI / CD step if assertions do not pass), then `rm -f /tmp/profiler_multi_josh.csv` before the next test step.
 - Add a new Python checker script at `.github/scripts/check_profiler_off.py` (modeled after `check_remote_step.py`). It should read the CSV and assert that every row's `avgHeightEvalDuration` value is less than 1 ms (i.e. `< 1.0`). Fail with a clear message if any row exceeds this threshold or if the file is missing. This script is shared with component 4.
 - Acceptance requires running the simulation and checker locally (outside CI) and confirming the expected behavior before considering this component complete.

### 2. Test with profiler local
We should again add to the "Test Additional Features" CI / CD checks. Specifically, we can use `examples/simulations/profiler_multi.josh` and, next to the earlier test from component 1, we should next run with `--enable-profiler` and we should ensure that the timings are reported but that there are non-zero values.

Details:
 - The simulation name is `ProfilerMultiExample`.
 - Run with `--replicates=1`.
 - The simulation writes to `/tmp/profiler_multi_josh.csv`. After running, immediately invoke the checker script, then `rm -f /tmp/profiler_multi_josh.csv`.
 - Add a new Python checker script at `.github/scripts/check_profiler_on.py` (modeled after `check_remote_step.py`). It should read the CSV and assert that at least one row's `avgHeightEvalDuration` value is `>= 1.0` ms. Fail with a clear message if no such row is found or if the file is missing. This script is shared with component 3.
 - We can expand the number of replicates or steps after trying locally if the timings are too small to trigger the threshold reliably.
 - Acceptance requires running the simulation and checker locally (outside CI) and confirming the expected behavior before considering this component complete.

### 3. Test remote with profiler
Similar to component 2, we should add to "Test Remote" CI / CD checks that uses `profiler_multi.josh` with the client run with `--enable-profiler`. We should ensure that there are non-zero values reported.

Details:
 - The simulation name is `ProfilerMultiExample`.
 - Add `--enable-profiler` as a CLI flag to `RunRemoteCommand`. When set, it should send `enableProfiler=true` as a new form field in the HTTP request to the worker. The server-side handler does not yet read this field — that is deferred to component 6.
 - Run with `--replicates=1`.
 - The output path `/tmp/profiler_multi_josh.csv` is the same as the local tests, but `testRemote` runs on a separate CI VM so there is no conflict.
 - Start the server normally (no profiler flag) using the same pattern as the existing `testRemote` job (background start, wait-for-ready loop, run test, stop with `pkill`).
 - Use the shared checker script `.github/scripts/check_profiler_on.py` (added in component 2). It asserts that at least one row's `avgHeightEvalDuration` is `>= 1.0` ms, failing with a clear message otherwise.
 - Run the checker immediately after the simulation, then `rm -f /tmp/profiler_multi_josh.csv`.
 - Note: this CI test is **intentionally RED** at the time of writing — the server does not yet honor the `enableProfiler` form field, so all timing values will be zero and the checker will fail. The validator agent should record this as an expected failure and confirm it is safe to continue.
 - Acceptance requires running the client locally (outside CI) with `--enable-profiler` and verifying the flag and form field are wired correctly, even though the full end-to-end behavior is not yet working.

### 4. Add test for profiler mismatch
Similar to component 3, we should again add to "Test Remote" CI / CD checks but we should have server run without `--enable-profiler`. At this stage, we should expect that there are zero values for this third case.

Details:
 - The simulation name is `ProfilerMultiExample`. Run with `--replicates=1`.
 - This is a "profiler mismatch" test: the client sends `--enable-profiler` (and the `enableProfiler=true` form field added in component 3) but the server does not yet honor that field. The result is zero timing values. Name the CI step clearly to reflect this — e.g. "Test profiler mismatch remote (client on, server unaware, expect zero)".
 - Reuse the server instance already running from component 3's steps — no need to stop and restart it. The `pkill` remains only at the end of the job.
 - Use the shared checker script `.github/scripts/check_profiler_off.py` (added in component 1). It asserts all rows have `avgHeightEvalDuration < 1.0` ms.
 - Run the checker immediately after the simulation, then `rm -f /tmp/profiler_multi_josh.csv`.
 - This test should be **GREEN** at time of writing — it documents the current broken behavior (mismatch returns zeros) and will remain green until component 5 flips it.
 - Acceptance requires running the client locally (outside CI) with `--enable-profiler` against a local server and confirming zero values are returned, before considering this component complete.

### 5. Switch to failing test
For the test added in component 4, we should change our expectations that we have non-zero values. This should fail ahead of component 6.

Details:
 - In the CI YAML, swap the checker for the component 4 step from `check_profiler_off.py` to `check_profiler_on.py`. No other changes needed.
 - This test is **intentionally RED** after this component. The validator agent should record this as an expected failure and confirm it is safe to continue.

### 6. Implement server-side profiler support
We should implement server-side handling of the `enableProfiler` per-request flag and add a server-global `--enable-profiler` CLI option to `ServerCommand`. After this component, the intentionally RED tests from components 3 and 5 should go GREEN.

Details:
 - In `JoshSimWorkerHandler.handleRequestTrusted()`, read the optional `enableProfiler` form field (sent by the client since component 3). If present and `"true"`, create `new ValueSupportFactory(favorBigDecimal, new TimedRecursiveValueResolverFactory())`. Otherwise use the existing `new ValueSupportFactory(favorBigDecimal)`. The existing constructor `ValueSupportFactory(boolean, ValueResolverFactory)` already supports this — no new factory mutation mechanism is needed.
 - Add `--enable-profiler` as a boolean CLI option to `ServerCommand` (default `false`). Pass it as a new constructor argument through `JoshSimServer` to `JoshSimWorkerHandler`. When the server-global flag is set, the handler forces profiling on for all requests regardless of the per-request form field.
 - Add unit tests to `JoshSimWorkerHandlerTest` (or equivalent) covering: (1) per-request profiler enabled via form field, (2) per-request profiler disabled, (3) server-global flag forces profiling regardless of form field.
 - After this component, no CI YAML changes are needed — the component 3 test already sends the form field and the component 5 test uses the same setup. Both should go GREEN once the handler reads the field correctly.
 - Note: a mutable `setValueResolverFactory` on `ValueSupportFactory` is not needed for this design. The per-request and server-global cases are both handled cleanly via constructor choice.
 - Note: after this component, the component 3 CI test and the component 5 CI test both verify the same scenario (client with `--enable-profiler`, server started without the global flag, expecting non-zero values via `check_profiler_on.py`). This is intentional — they arrived at the same end state through different steps of the TDD cycle and together give confidence that per-request profiling works correctly.

### 7. Integration test for server-global profiler flag
We should add a CI integration test to "Test Remote" that verifies the server-global `--enable-profiler` flag works end-to-end, independently of the per-request form field. This component was not needed to validate component 6's per-request path but is required to confirm the server-global path introduced in component 6.

Details:
 - The simulation name is `ProfilerMultiExample`. Run with `--replicates=1`.
 - Add a new step to the `testRemote` CI job that starts the server WITH `--enable-profiler` (i.e. `java -jar joshsim.jar server --enable-profiler`), then runs the client WITHOUT `--enable-profiler` (no per-request flag, no `enableProfiler` form field sent). This isolates the server-global flag as the only source of profiling.
 - Use the shared checker script `.github/scripts/check_profiler_on.py`. It should find at least one row with `avgHeightEvalDuration >= 1.0` ms, confirming the server-global flag forced profiling even without a per-request flag from the client.
 - Run the checker immediately after the simulation, then `rm -f /tmp/profiler_multi_josh.csv`.
 - This server instance is separate from the one used in components 3–5 (which runs without `--enable-profiler`). Stop the previous server with `pkill` before starting this one, and stop this one at the end of the job.
 - This test should be **GREEN** at time of writing since component 6 implements the server-global flag.
 - Acceptance requires running the server locally with `--enable-profiler` and the client without it, confirming non-zero values before considering this component complete.

### 8. Add multi-replicate case
We should again add to the "Test Additional Features" CI / CD checks. Specifically, we can use `examples/simulations/profiler_multi.josh` and, next to the earlier test from components 2 and 3, we should next run with `--enable-profiler` and we should ensure that the timings are reported but that there are non-zero values. However, we should do 5 replicates this time.

## Notes
Co-prepared with Claude (Anthropic). Original written manually, refined iteratively with human in loop, and edited afterwards manually.

sampottinger:
Closes #391. Specifically, we expand CI / CD to cover profiler in live executions and with remote tests. We also add support for Josh Cloud or self-run remotes to enable the profiler based on user request. Moves some longer inline code snippets from YAML to `sh` files for readability.


.github/scripts/check_profiler_on.py
```diff
@@ -0,0 +1,46 @@
+"""Check that profiler timings are non-trivial (> 0.1 ms, profiler enabled).
```
sampottinger: No need to mention any components by number in code. That idea goes away after the task is complete. Let’s also try to merge `check_profiler_on.py` and `check_profiler_off.py` to just a `check_profiler.py` with an argument that switches between behaviors.

.github/workflows/build.yaml
```diff
@@ -188,6 +188,57 @@ jobs:
             --replicates 5
       - name: Check step labeling
         run: python3 .github/scripts/check_remote_step.py
+      - name: Test profiler on remote (client --enable-profiler, expect non-zero timings)
+        run: |
+          java -Xmx2g -jar build/libs/joshsim-fat.jar runRemote \
```
sampottinger: Can this go into a shell script in the github actions `scripts` directory (already exists)?

.github/workflows/build.yaml
```diff
@@ -188,6 +188,57 @@ jobs:
...
+            ProfilerMultiExample \
+            --endpoint http://localhost:8085/runReplicates \
+            --api-key testkey \
+            --replicates=1 \
+            --enable-profiler
+          python3 .github/scripts/check_profiler_on.py
+          rm -f /tmp/profiler_multi_josh.csv
+      - name: Test profiler mismatch remote (client --enable-profiler, server honors field, expect non-zero timings)
+        run: |
+          java -Xmx2g -jar build/libs/joshsim-fat.jar runRemote \
```
sampottinger: Can this go into a shell script in the github actions `scripts` directory (already exists)?

.github/workflows/build.yaml
```diff
@@ -188,6 +188,57 @@ jobs:
...
+            --enable-profiler
+          python3 .github/scripts/check_profiler_on.py
+          rm -f /tmp/profiler_multi_josh.csv
+      - name: Stop server before profiler server
+        run: pkill -f joshsim-fat.jar || true
+      - name: Start profiler server
+        run: java -Xmx2g -jar build/libs/joshsim-fat.jar server --enable-profiler > /tmp/josh_profiler_server.log 2>&1 &
+      - name: Wait for profiler server ready
+        run: |
+          for i in {1..30}; do
```
sampottinger: Can this go into a shell script in the github actions `scripts` directory (already exists)?

.github/workflows/build.yaml
```diff
@@ -188,6 +188,57 @@ jobs:
...
+            sleep 1
+            if [ $i -eq 30 ]; then
+              echo "Profiler server failed to start. Log:"
+              cat /tmp/josh_profiler_server.log
+              exit 1
+            fi
+          done
+      - name: Test server-global profiler flag (client without --enable-profiler)
+        run: |
+          java -Xmx2g -jar build/libs/joshsim-fat.jar runRemote \
```
sampottinger: Can this go into a shell script in the github actions `scripts` directory (already exists)?

.github/workflows/build.yaml
```diff
@@ -368,6 +419,32 @@ jobs:
         run: bash examples/test_data_option.sh
       - name: Test job configuration template functionality
         run: bash examples/test_job_config.sh
+      - name: Test profiler off local (expect zero timings)
+        run: |
+          java -Xmx2g -jar build/libs/joshsim-fat.jar run \
```
sampottinger: Can this go into a shell script in the github actions `scripts` directory (already exists)?

.github/workflows/build.yaml
```diff
@@ -368,6 +419,32 @@ jobs:
...
+            --replicates=1 \
+            examples/simulations/profiler_multi.josh \
+            ProfilerMultiExample
+          python3 .github/scripts/check_profiler_off.py
+          rm -f /tmp/profiler_multi_josh.csv
+      - name: Test profiler on local (expect non-zero timings)
+        run: |
+          java -Xmx2g -jar build/libs/joshsim-fat.jar run \
```
sampottinger: Can this go into a shell script in the github actions `scripts` directory (already exists)?

.github/workflows/build.yaml
```diff
@@ -368,6 +419,32 @@ jobs:
...
+            --replicates=1 \
+            --enable-profiler \
+            examples/simulations/profiler_multi.josh \
+            ProfilerMultiExample
+          python3 .github/scripts/check_profiler_on.py
+          rm -f /tmp/profiler_multi_josh.csv
+      - name: Test profiler on local with 5 replicates (expect non-zero timings)
+        run: |
+          java -Xmx2g -jar build/libs/joshsim-fat.jar run \
```
sampottinger: Can this go into a shell script in the github actions `scripts` directory (already exists)?

.github/workflows/build.yaml
sampottinger: Can this go into a shell script in the github actions `scripts` directory (already exists)?

sampottinger: Can this go into a shell script in the github actions `scripts` directory (already exists)?

sampottinger: Can this go into a shell script in the github actions `scripts` directory (already exists)?

sampottinger: Can this go into a shell script in the github actions `scripts` directory (already exists)?

sampottinger: Can this go into a shell script in the github actions `scripts` directory (already exists)?

src/main/java/org/joshsim/cloud/JoshSimServer.java
```diff
@@ -114,7 +116,8 @@ public JoshSimServer(CloudApiDataLayer dataLayer, boolean useHttp2, String worke
         )
         .addPrefixPath(
             "/runReplicate",
-            new JoshSimWorkerHandler(dataLayer, true, Optional.empty(), serialPatches)
+            new JoshSimWorkerHandler(
+                dataLayer, true, Optional.empty(), serialPatches, enableProfiler)
```
sampottinger: Can we make this in a local variable above and pass as a single argument without line continuation?

src/main/java/org/joshsim/cloud/JoshSimWorkerHandler.java
```diff
@@ -152,6 +157,10 @@ public Optional<String> handleRequestTrusted(HttpServerExchange httpServerExchan
         ? formData.getFirst("outputSteps").getValue() : "";
     final Optional<Set<Integer>> outputSteps = parseOutputSteps(outputStepsStr);
 
+    boolean requestEnableProfiler = formData.contains("enableProfiler")
```
sampottinger: Can we pass to a private method with JavaDoc that takes in `formData` called `getProfilerEnabled` returning Boolean that has an if statement that returns early if form data does not contain `enableProfiler`?

src/main/java/org/joshsim/cloud/JoshSimWorkerHandler.java
```diff
@@ -189,6 +198,21 @@ public Optional<String> handleRequestTrusted(HttpServerExchange httpServerExchan
...
+   * Build a ValueSupportFactory configured for optional profiling.
+   *
+   * @param favorBigDecimal Flag indicating if decimal values should favor BigDecimal.
+   * @param useProfiler If true, the factory will use TimedRecursiveValueResolverFactory so that
+   *     evaluation durations are captured; otherwise uses the default
+   *     RecursiveValueResolverFactory.
+   * @return A ValueSupportFactory configured according to the given flags.
+   */
+  ValueSupportFactory buildValueFactory(boolean favorBigDecimal, boolean useProfiler) {
+    return useProfiler
```
sampottinger: This is long enough let’s make into a if / else with curly braces.

src/main/java/org/joshsim/pipeline/remote/RunRemoteContextBuilder.java
```diff
@@ -56,6 +56,9 @@ public class RunRemoteContextBuilder {
   private int maxConcurrentWorkers = 10;
   private int replicateNumber = 0;
 
+  // Profiler flag
```
sampottinger: Don’t need this inline comment… it’s obvious from the code itself.

src/main/java/org/joshsim/pipeline/remote/RunRemoteContextBuilder.java
```diff
@@ -226,7 +240,7 @@ public RunRemoteContext build() {
         joshCode.get(), externalDataSerialized.get(),
         metadata.get(), progressCalculator.get(),
         outputOptions.get(), minioOptions.get(),
-        maxConcurrentWorkers
+        maxConcurrentWorkers, enableProfiler
```
sampottinger: Let’s put `enableProfiler` on its own line for consistency with prior code.

sampottinger (commented):
Some clean up. When making the scripts in the GitHub actions directory, please look for places where scripts can be consolidated where args may be passed from the call in the GitHub actions YAML.

.github/scripts/check_profiler_on.py
sampottinger: Done — merged `check_profiler_on.py` and `check_profiler_off.py` into a single `check_profiler.py` that accepts `on` or `off` as a positional argument. All 6 call sites in `build.yaml` updated accordingly; the two old scripts deleted. Component number references removed from the docstring. Committed in `8ac2b254`. (Aided by Claude)

.github/workflows/build.yaml
sampottinger: Done — extracted into a shell script under `.github/scripts/`. Reusable scripts accept arguments so multiple YAML steps share the same script (e.g. `run_profiler_local.sh`, `run_profiler_remote.sh`, `wait_for_server.sh`). YAML steps now call `bash .github/scripts/<script>.sh [args]` as a single line. Committed in `cd1c583f`. (Aided by Claude)

.github/workflows/build.yaml
sampottinger: Done — extracted into a shell script under `.github/scripts/`. Reusable scripts accept arguments so multiple YAML steps share the same script (e.g. `run_profiler_local.sh`, `run_profiler_remote.sh`, `wait_for_server.sh`). YAML steps now call `bash .github/scripts/<script>.sh [args]` as a single line. Committed in `cd1c583f`. (Aided by Claude)

.github/workflows/build.yaml
sampottinger: Done — extracted into a shell script under `.github/scripts/`. Reusable scripts accept arguments so multiple YAML steps share the same script (e.g. `run_profiler_local.sh`, `run_profiler_remote.sh`, `wait_for_server.sh`). YAML steps now call `bash .github/scripts/<script>.sh [args]` as a single line. Committed in `cd1c583f`. (Aided by Claude)

.github/workflows/build.yaml
sampottinger: Done — extracted into a shell script under `.github/scripts/`. Reusable scripts accept arguments so multiple YAML steps share the same script (e.g. `run_profiler_local.sh`, `run_profiler_remote.sh`, `wait_for_server.sh`). YAML steps now call `bash .github/scripts/<script>.sh [args]` as a single line. Committed in `cd1c583f`. (Aided by Claude)

.github/workflows/build.yaml
sampottinger: Done — extracted into a shell script under `.github/scripts/`. Reusable scripts accept arguments so multiple YAML steps share the same script (e.g. `run_profiler_local.sh`, `run_profiler_remote.sh`, `wait_for_server.sh`). YAML steps now call `bash .github/scripts/<script>.sh [args]` as a single line. Committed in `cd1c583f`. (Aided by Claude)

.github/workflows/build.yaml
sampottinger: Done — extracted into a shell script under `.github/scripts/`. Reusable scripts accept arguments so multiple YAML steps share the same script (e.g. `run_profiler_local.sh`, `run_profiler_remote.sh`, `wait_for_server.sh`). YAML steps now call `bash .github/scripts/<script>.sh [args]` as a single line. Committed in `cd1c583f`. (Aided by Claude)

.github/workflows/build.yaml
sampottinger: Done — extracted into a shell script under `.github/scripts/`. Reusable scripts accept arguments so multiple YAML steps share the same script (e.g. `run_profiler_local.sh`, `run_profiler_remote.sh`, `wait_for_server.sh`). YAML steps now call `bash .github/scripts/<script>.sh [args]` as a single line. Committed in `cd1c583f`. (Aided by Claude)

.github/workflows/build.yaml
sampottinger: Done — extracted into a shell script under `.github/scripts/`. Reusable scripts accept arguments so multiple YAML steps share the same script (e.g. `run_profiler_local.sh`, `run_profiler_remote.sh`, `wait_for_server.sh`). YAML steps now call `bash .github/scripts/<script>.sh [args]` as a single line. Committed in `cd1c583f`. (Aided by Claude)

sampottinger: Done — extracted into a shell script under `.github/scripts/`. Reusable scripts accept arguments so multiple YAML steps share the same script (e.g. `run_profiler_local.sh`, `run_profiler_remote.sh`, `wait_for_server.sh`). YAML steps now call `bash .github/scripts/<script>.sh [args]` as a single line. Committed in `cd1c583f`. (Aided by Claude)

sampottinger: Done — extracted into a shell script under `.github/scripts/`. Reusable scripts accept arguments so multiple YAML steps share the same script (e.g. `run_profiler_local.sh`, `run_profiler_remote.sh`, `wait_for_server.sh`). YAML steps now call `bash .github/scripts/<script>.sh [args]` as a single line. Committed in `cd1c583f`. (Aided by Claude)

sampottinger: Done — extracted into a shell script under `.github/scripts/`. Reusable scripts accept arguments so multiple YAML steps share the same script (e.g. `run_profiler_local.sh`, `run_profiler_remote.sh`, `wait_for_server.sh`). YAML steps now call `bash .github/scripts/<script>.sh [args]` as a single line. Committed in `cd1c583f`. (Aided by Claude)

sampottinger: Done — extracted into a shell script under `.github/scripts/`. Reusable scripts accept arguments so multiple YAML steps share the same script (e.g. `run_profiler_local.sh`, `run_profiler_remote.sh`, `wait_for_server.sh`). YAML steps now call `bash .github/scripts/<script>.sh [args]` as a single line. Committed in `cd1c583f`. (Aided by Claude)

src/main/java/org/joshsim/cloud/JoshSimServer.java
sampottinger: Done — `JoshSimWorkerHandler` is now constructed into a `workerHandler` local variable before the path-handler chain, then passed as a single identifier to `addPrefixPath`. Committed in `70932ad9`. (Aided by Claude)

src/main/java/org/joshsim/cloud/JoshSimWorkerHandler.java
sampottinger: Done — extracted into a private `getProfilerEnabled(FormData formData)` method with JavaDoc. Returns `false` early if the field is absent, then returns `"true".equals(...)` otherwise. `handleRequestTrusted` now calls it directly. Committed in `70932ad9`. (Aided by Claude)

src/main/java/org/joshsim/cloud/JoshSimWorkerHandler.java
sampottinger: Done — converted to a standard if/else block with curly braces on both branches. Committed in `70932ad9`. (Aided by Claude)

src/main/java/org/joshsim/pipeline/remote/RunRemoteContextBuilder.java
sampottinger: Done — removed the `// Profiler flag` comment. Committed in `70932ad9`. (Aided by Claude)

src/main/java/org/joshsim/pipeline/remote/RunRemoteContextBuilder.java
sampottinger: Done — `enableProfiler` is now on its own line, consistent with the other arguments in the `build()` constructor call. Committed in `70932ad9`. (Aided by Claude)

src/main/java/org/joshsim/cloud/JoshSimWorkerHandler.java
```diff
@@ -201,6 +225,19 @@ private static Optional<Set<Integer>> parseOutputSteps(String outputSteps) {
...
+  /**
+   * Determine whether the per-request enableProfiler form field is set to true.
+   *
+   * @param formData The form data submitted with the request.
+   * @return True if the enableProfiler field is present and equal to "true", false otherwise.
+   */
+  private boolean getProfilerEnabled(FormData formData) {
+    if (!formData.contains("enableProfiler")) {
+      return false;
+    }
```
sampottinger, line 241: Let’s do an if / else here and let’s put `formData.getFirst("enableProfiler").getValue()` in a local var inside the else.

src/main/java/org/joshsim/cloud/JoshSimWorkerHandler.java
sampottinger: Let’s do a `getOutputSteps` like our `getProfilerEnabled`.

sampottinger (commented):
Great! Just two final pieces.

src/main/java/org/joshsim/cloud/JoshSimWorkerHandler.java
sampottinger: Done — `getProfilerEnabled` now uses if/else with `formData.getFirst("enableProfiler").getValue()` extracted to a local variable `value` inside the else branch. Committed in `b60b6aa5`. (Aided by Claude)

src/main/java/org/joshsim/cloud/JoshSimWorkerHandler.java
sampottinger: Done — extracted into a `getOutputSteps(FormData formData)` private method with JavaDoc, using if/else matching the `getProfilerEnabled` pattern. `handleRequestTrusted` now calls it directly. Committed in `b60b6aa5`. (Aided by Claude)

sampottinger:
I think this is ready for your review @GondekNP!

GondekNP (commented):
Nice work @sampottinger !

PR #397, opened 2026-04-14
Fix MinIO streaming silent failure for large uploads

### LINKED ISSUE #396: MinIO streaming silently fails for large runs (>10MB)
(Claude Summary)

## Bug Summary

Large batch simulations complete with `"status": "complete"` but no CSV data appears in MinIO. Small runs work fine. Two bugs combine to cause this:

1. **`JvmQueueService.join()` silently swallows exceptions** from the worker thread, making MinIO upload failures invisible.
2. **`MinioOutputStreamStrategy.uploadWithRetry()` retries on an already-consumed stream**, which either hangs or produces empty uploads.

### Bug 1: Exception swallowing in `JvmQueueService.join()`

The worker thread runs inside `executorService.submit(() -> { ... callback.onEnd(); })`. If `onEnd()` throws (which is where MinIO upload completion happens via `outputStream.close()`), the exception is captured in the `Future` returned by `submit()` — but `join()` never calls `future.get()`:

```java
// JvmQueueService.java
public void join() {
    active.set(false);
    executorService.shutdown();
    while (!executorService.isTerminated()) {
        trySleep();  // never retrieves the exception from the Future
    }
}
```

This means the batch handler proceeds to write `"status": "complete"` even when the upload failed.

### Bug 2: Broken retry on consumed `BlockingQueueInputStream`

`uploadWithRetry()` retries `putObject` on the same `BlockingQueueInputStream`. Since `ArrayBlockingQueue.take()` is destructive, data consumed during the first attempt is gone. On retry, the stream either:
- **Blocks forever** waiting for data that was already consumed
- **Uploads corrupt/empty data** from wherever the stream happens to be

The code comment says *"the queue can be read multiple times if needed"* — this is incorrect for a destructive queue.

For small runs (< 10MB), `putObject` uses a single PUT that succeeds quickly. For large runs (> 10MB), the MinIO SDK switches to **multipart upload** — a longer-lived operation more susceptible to transient failures. One failure triggers the broken retry, and the result is hidden by Bug 1.

## Why existing tests don't catch this

**Unit tests:**
- `JvmQueueServiceTest` — test callbacks never throw, so the swallowed-exception path is untested
- `MinioOutputStreamStrategyTest` — retry test uses a Mockito mock that returns instantly without consuming the stream, so it never exercises the real broken behavior

**CI integration tests (`test-minio.yaml`):**
- All test simulations produce trivially small data (< 1KB), well under the 10MB multipart threshold
- All tests run on localhost, so transient network failures never occur
- `csv_export.josh`: 1 patch x 10 steps x 2 values = ~20 CSV rows
- `backpressure.josh`: 1 patch x 10 steps x 1 value = ~10 rows (exports `mean()`, not per-organism)

## Proposed Fix

1. **`JvmQueueService.join()`**: Store the `Future<?>` from `submit()`, call `future.get()` in `join()` to propagate worker thread exceptions.
2. **`MinioOutputStreamStrategy`**: Remove the broken retry loop. Streaming uploads cannot be retried at this level — retry belongs at a higher level (e.g., re-run the simulation).
3. **Tests**: Add unit tests that reproduce both bugs (fail-first), plus a CI integration test with > 10MB data to exercise the multipart upload path.

## Files Involved

- `src/main/java/org/joshsim/compat/JvmQueueService.java`
- `src/main/java/org/joshsim/lang/io/MinioOutputStreamStrategy.java`
- `src/test/java/org/joshsim/compat/JvmQueueServiceTest.java`
- `src/test/java/org/joshsim/lang/io/MinioOutputStreamStrategyTest.java`
- `.github/workflows/test-minio.yaml`

GondekNP:
## Summary

~~Fixes #396. Large batch simulations complete with `"status": "complete"` but no CSV data appears in MinIO. Small runs work fine.~~

See comment below!

- **`JvmQueueService.join()` now propagates worker thread exceptions** instead of silently swallowing them. The `Future` from `submit()` is stored and `get()` is called in `join()` to surface errors from `onEnd()` (where MinIO upload completion happens).
- **Removed broken retry logic from `MinioOutputStreamStrategy`**. The retry loop re-read an already-consumed `BlockingQueueInputStream` — data taken from an `ArrayBlockingQueue` is gone, so retries would either block forever or upload empty/corrupt data. Streaming uploads cannot be retried at this level; a single-attempt upload with clean error propagation is correct.
- **Added CI integration test** with a >10MB simulation to exercise the MinIO SDK multipart upload path, which was the boundary where failures occurred.

## Test plan

- [x] New unit test: `JvmQueueServiceTest.testExceptionInOnEndPropagatesToJoin` — verified it fails before fix, passes after
- [x] New unit test: `JvmQueueServiceTest.testExceptionInOnTaskPropagatesToJoin` — same
- [x] New unit test: `MinioOutputStreamStrategyTest.testUploadFailure_afterStreamConsumed_propagatesError` — verified it fails before fix (retry silently "succeeds" with empty data), passes after
- [x] Updated `testUploadFailure_propagatesAsIoException` — replaces old retry-exhaustion test
- [x] All existing tests pass (`./gradlew test` — 0 failures)
- [x] CI: new `large_export.josh` test step in `test-minio.yaml` verifies >10MB multipart upload succeeds

### DIFF
--- a/.github/workflows/test-minio.yaml
+++ b/.github/workflows/test-minio.yaml
@@ -200,6 +200,36 @@ jobs:
 
           echo "✓ Queue backpressure test passed"
 
+      - name: Test large payload upload (multipart)
+        run: |
+          echo "Testing large payload upload to exercise multipart upload path..."
+
+          # Run simulation that generates >10MB of entity-level CSV data
+          # 1500 organisms * 100 steps * 3 fields = ~15MB CSV
+          java -Xmx2g -jar build/libs/joshsim-fat.jar run \
+            --replicates=1 \
+            examples/test/minio/large_payload.josh \
+            LargePayloadTest \
+            --minio-endpoint=http://localhost:9000 \
+            --minio-access-key=minioadmin \
+            --minio-secret-key=minioadmin \
+            --export-queue-size=10000 \
+            --output-steps=100
+
+          # Verify file was uploaded
+          ./mc stat testminio/josh-test-bucket/results/large_payload_0.csv
+
+          # Verify file size exceeds 10MB multipart threshold
+          SIZE_BYTES=$(./mc stat testminio/josh-test-bucket/results/large_payload_0.csv --json | grep -o '"size":[0-9]*' | cut -d':' -f2)
+          echo "  Large payload size: $SIZE_BYTES bytes"
+
+          if [ "$SIZE_BYTES" -gt 10485760 ]; then
+            echo "✓ Large payload test passed (${SIZE_BYTES} bytes > 10MB multipart threshold)"
+          else
+            echo "✗ Large payload test failed: expected >10MB but got ${SIZE_BYTES} bytes"
+            exit 7
+          fi
+
       - name: Test error handling - missing credentials
         run: |
           echo "Testing error handling with missing credentials..."
@@ -341,6 +371,7 @@ jobs:
           echo "✓ File upload verification"
           echo "✓ Large CSV export (>10MB multipart upload)"
           echo "✓ Queue backpressure handling"
+          echo "✓ Large payload multipart upload"
           echo "✓ Error handling validation"
           echo "✓ MinioPollingStrategy integration"
           echo "✓ batchRemote end-to-end (server + MinIO)"
--- a/.github/workflows/test-minio.yaml
+++ b/.github/workflows/test-minio.yaml
@@ -160,6 +160,26 @@ jobs:
             exit 5
           fi
 
+      - name: Test large CSV export to MinIO (multipart upload)
+        run: |
+          echo "Testing large export forcing multipart upload (>10MB)..."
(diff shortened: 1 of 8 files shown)


GondekNP:
For now this closes #396. However - the issue wasn't as expected, in fact I can't reproduce a case where, with the right setup, a large file fails to make it to minio. This was primarily my misunderstanding of how the streaming operates, while we are passing along chunks during a streaming operation of a csv to mino, it's not like appending to a csv on disk, it's an all or nothing deal (either it works or it doesn't).

Due to the fact that this manifested late at night with a big OOM job in the first place, I don't think a job ever actually finished AND failed to make it to minio. This is kind of annoying for other reasons (see #403 - all-or-nothing particuarly sucks in the case of unexpected failures where partial results are still useful, especially as a diagnostic), but doesn't seem to be a bug based on expected behavior.

PR #411, opened 2026-04-17
Add CI workflow to build and push batch worker image to GHCR

GondekNP:
## Summary
- New workflow `.github/workflows/build-batch-image.yaml` that builds the `joshsim-batch` Docker image and pushes to `ghcr.io/schmidtdse/josh/joshsim-batch`
- Single image supports both run and preprocess — `Dockerfile.batch` includes both `run-entrypoint.sh` and `preprocess-entrypoint.sh`
- Triggers on push to `main`, `dev`, `feat/k8s-batch` when cloud-img/, src/main/, or build.gradle change
- Supports `workflow_dispatch` for manual trigger (needed for first push)
- Tags: `<sha>` (immutable), `<branch>` (mutable), `latest` (main/dev only)
- Uses `GITHUB_TOKEN` for GHCR auth — no additional secrets needed

## Test plan
- [ ] Merge and trigger manually via `workflow_dispatch`
- [ ] Image appears at `ghcr.io/schmidtdse/josh/joshsim-batch`
- [ ] Make package public (Settings → Packages → Change visibility)
- [ ] GKE pods can pull the image (resolves ErrImagePull from integration testing)

### DIFF
--- a/.github/workflows/build-batch-image.yaml
+++ b/.github/workflows/build-batch-image.yaml
@@ -1,0 +1,91 @@
+name: Build Batch Worker Image
+
+on:
+  push:
+    branches: [main, dev, feat/k8s-batch]
+    paths:
+      - 'cloud-img/**'
+      - 'src/main/**'
+      - 'build.gradle'
+  workflow_dispatch:
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: schmidtdse/josh/joshsim-batch
+
+jobs:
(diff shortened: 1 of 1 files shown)


GondekNP (commented):
Actually, this whole workflow should be running as part of .github/build.yaml (at the end, taking the built fatjar from the proper step) - we don't want to build and deploy in one step without handoff for hardening reasons

PR #413, opened 2026-04-17
Add --replicate-index flag to RunCommand

GondekNP:
## Summary
Add `--replicate-index` flag to `RunCommand`, mutually exclusive with `--replicates`. Runs a single replicate at the specified index so `{replicate}` template paths resolve correctly.

Used by K8s indexed Jobs where each pod runs one replicate — pod 3 uses `--replicate-index=3`, producing `smoke_3.csv` instead of `smoke_0.csv`.

```bash
# Old (all pods write replicate 0):
java -jar joshsim.jar run sim.josh Main --replicates=1

# New (pod writes at its assigned index):
java -jar joshsim.jar run sim.josh Main --replicate-index=$JOB_COMPLETION_INDEX
```

## Test plan
- [x] `./gradlew test checkstyleMain` — all pass
- [x] Mutually exclusive validation: `--replicate-index=3 --replicates=5` → error
- [x] Negative index validation: `--replicate-index=-1` → error
- [x] Existing `--replicates` behavior unchanged

### DIFF
--- a/src/main/java/org/joshsim/command/RunCommand.java
+++ b/src/main/java/org/joshsim/command/RunCommand.java
@@ -368,65 +368,47 @@ public class RunCommand implements Callable<Integer> {
     boolean sizeMetersFull = sizeStr.equals("meter") || sizeStr.equals("meters");
     boolean sizeMeters = sizeMetersFull || sizeMeterAbbreviated;
 
-    // Single replicate mode: run one replicate at the specified index and return.
-    if (replicateIndex != null) {
-      JoshJob currentJob = jobs.get(0);
-      if (!currentJob.getFilePaths().isEmpty()) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/joshsim/command/RunCommand.java
```diff
@@ -357,7 +380,11 @@ public Integer call() {
         inputStrategy = new JvmMappedInputGetter(currentJob.getFilePaths());
       }
 
-      for (int currentReplicate = 0; currentReplicate < currentJob.getReplicates();
+      int startReplicate = replicateIndex != null ? replicateIndex : 0;
+      int endReplicate = replicateIndex != null
+          ? replicateIndex + 1 : currentJob.getReplicates();
```
GondekNP: This is kind of an ugly way to run a single replicate, we should factor out and early return

Issue #419, opened 2026-04-21
Cold-start spin-up for external data (random temporal resampling)


GondekNP:
## Motivation

Ecological simulations often need a **spin-up period** to bring the modeled ecosystem to a quasi-equilibrium state before the "real" simulation begins with year-aligned external data. Without spin-up, initial conditions (e.g., forest age structure, soil moisture) are artificial and can dominate early results.

## Proposed Feature

Add support for **cold-start spin-up** of external `.jshd` data. During the cold-start period, for each simulation year, Josh would randomly sample one of the available timesteps from the original `.jshd` time series (ignoring the source year) and use that entire spatial grid. After the cold-start period ends, the simulation switches to the real time-aligned data.

### Example use case

A user has 50 years of climate data preprocessed into `temperature.jshd` (timesteps 0–49). They want to run 100 years of spin-up followed by the 50-year "real" simulation:

- **Years 0–99 (cold start):** Each year, the engine randomly picks one of timesteps 0–49 from the jshd and uses that grid. Each replicate gets a different random draw.
- **Years 100–149 (real simulation):** Year 100 maps to jshd timestep 0, year 101 to timestep 1, etc.

This produces 150 total simulation years where the first 100 bring the ecosystem to equilibrium.

## Current Architecture

The external data pipeline currently works as:

1. `external temperatureGrid` in Josh → `JoshExternalVisitor` compiles to `machine.pushExternal(name, stepCount)`
2. `SingleThreadEventHandlerMachine.pushExternal()` → `bridge.getExternal(geoKey, name, step)`
3. `MinimalEngineBridge.getExternal()` → `layer.getAt(key, step)` on the cached `DataGridLayer`
4. `DoublePrecomputedGrid.getAt(x, y, timestep)` does direct index lookup: `timestepCut = timestep - minTimestep`

Steps beyond the jshd's timestep range throw `IllegalArgumentException`. There is no concept of timestep remapping.

## Design Considerations

- **Runtime, not preprocess-time:** The random sampling must happen per-replicate so each replicate gets different draws. This rules out a pure preprocess-time solution.
- **Spatial coherence:** For a given cold-start year, the same source timestep should be used across all grid cells (sample a whole year, not cell-by-cell).
- **Reproducibility:** Random draws should be seeded so results are reproducible given the same replicate index.
- **Minimal language changes:** Ideally this could be configured via the simulation stanza or CLI flags rather than requiring new grammar.

## Acceptance Criteria

[...]

Issue #404, opened 2026-04-15
Progress bar reports wrong total steps (11 instead of 86)


GondekNP:
## Bug

The progress bar reports `step N/11` regardless of the actual `stepsHigh` value in the config. The simulation itself runs the correct number of steps (86 in this case), but the progress denominator is wrong, resulting in percentages like `781.8%`.

## Observed behavior

```
Progress: 0.0% (step 0/11)
Progress: 9.1% (step 1/11)
...
Progress: 781.8% (step 86/11)
```

## Expected behavior

```
Progress: 0.0% (step 0/86)
Progress: 1.2% (step 1/86)
...
Progress: 100.0% (step 86/86)
```

## Config

```
stepsHigh = 86 count
```

The simulation completes all 86 steps correctly — only the progress reporting is wrong.

## Bottle

Bottle attached: `bottle_test-binom-dev_b7de10c4d6bc_20260415_001805.tar.gz`

(Note: bottle is too large to attach here — available on request or at the josh-models repo)

## Environment

- Josh JAR: DEV branch
- JVM: OpenJDK 25.0.2
- joshpy with `stream_output=True`

GondekNP:
Related observation from #419 (cold-start work): with `--cold-start N`, the progress denominator is wrong in a different way — it shows negative percentages because the engine reports `completedStep` in the user's declared step range while the loop runs extra steps before `steps.low`. Example output:

```
Progress: -33.3% (step -2/6)
Progress: -16.7% (step -1/6)
Progress: 0.0% (step 0/6)
Progress: 16.7% (step 1/6)
Progress: 33.3% (step 2/6)
Progress: 50.0% (step 3/6)
```

Same root cause as this issue: `SimulationMetadataExtractor` / `ProgressCalculator` doesn't account for the actual range of `completedStep` values the loop will produce (config-driven `stepsHigh`, cold-start prepended steps, etc.). Recording here as supporting evidence; not fixing as part of #419.

PR #424, opened 2026-04-22
Add cold-start spin-up for external data (#419)

GondekNP:
## Summary
- Adds `--cold-start N` flag that prepends N warm-up steps before the real simulation
- During cold-start, external `.jshd` data timesteps are randomly sampled (per-replicate, reproducible with `--seed`); after cold-start, data is time-aligned
- Export is suppressed during cold-start steps — outputs look identical to a no-cold-start baseline for the real simulation period
- Propagated through the full execution chain: local run, batchRemote CLI, HTTP target (form field), K8s target (env var)

Closes SchmidtDSE/josh#419.

## Implementation

**Decorator + bridge step shift, no grammar changes:**

1. **`ColdStartDataGridLayer`** (new) — decorator over `DataGridLayer`:
   - Timesteps `[0, N)` → randomly pre-generated sample from inner grid's range
   - Timesteps `[N, N + innerCount)` → offset to inner grid's real timesteps
   - Anything else (e.g., literal `external X at 2020`) → passthrough to inner
2. **`MinimalEngineBridge`** — initializes `currentStep` at `steps.low - coldStartDuration`, so the main loop naturally runs the extra iterations. `absoluteStep` still starts at 0, which is what the decorator keys off. Grid wrapping happens inside `computeIfAbsent` so each replicate gets its own `Random` derived from `SharedRandom.get().nextLong()` (seeded when `--seed` is set).
3. **`JoshSimFacadeUtil`** — main loop skips export when `completedStep < bridge.getStartTimestep()`. Wraps the `IncrementalPatchExportCallback` during cold-start so patches still get frozen (for `prior`-state access) but aren't written to output.
4. **CLI & remote plumbing** — `--cold-start` on `RunCommand` and `BatchRemoteCommand`; `coldStart` form field on `/runBatch`; `JOSH_COLD_START` env var on K8s pods; passthrough through `BatchJobStrategy` and `RemoteBatchTarget.dispatch()`.

## Verified end-to-end

Ran the Joshua Tree canonical bottle (`joshpy-bottles/extracted/bottle_866ceb457fca`) with and without `--cold-start 2`. Row count and simulation behavior match the no-cold-start baseline (2025 rows from 4 real steps); cold-start steps run but produce no export output. `meta.year` returns the engine's `currentStep` value — naturally negative during cold-start, crosses zero at the real-simulation boundary.

Known interaction documented in `llms-full.txt`: if a script defines its own `year` attribute (e.g., `year.init = 0; year.step = prior.year + 1`), it shadows the built-in `meta.year`, so exports will show accumulated values through cold-start. Recommendation is to use `meta.year` directly.

## Tests
- New unit tests for `ColdStartDataGridLayer` (11 cases: remapping, determinism with seed, passthrough, edge cases)
- Updated constructor-signature-affected tests in `MinimalEngineBridgeTest`, `QueryCacheEngineBridgeTest`, `MinimalEngineBridgeConcurrencyTest`, `BatchJobStrategyTest`, `HttpBatchTargetTest`, `KubernetesTargetTest`
- `./gradlew test`: 1946 tests, 0 failures

## Follow-ups (not in this PR)
- Progress bar shows negative percentages during cold-start (e.g., `-33.3% (step -2/6)`). Added as evidence to existing SchmidtDSE/josh#404 — same root cause (metadata extraction doesn't account for dynamic step ranges).

## Test plan
- [ ] CI passes on this branch
- [ ] Run the bottle with `--cold-start 2` locally; verify output matches the baseline for the real-simulation rows
- [ ] Dispatch a batchRemote job with `--cold-start` to confirm the flag propagates through HTTP/K8s targets
- [ ] Confirm `meta.year` behavior on a script that does NOT define its own `year` attribute


GondekNP:
Stale from before the k8s merge - going to close now, and if we decide to revive we can deal with the conflicts later on.

PR #427, opened 2026-04-23
Add .jshdz format: XZ-compressed JSHD via decorator pattern

### LINKED ISSUE #425: Jshd -> Jsdz (jshd efficiency / compression)
Co-created by me and @sampottinger !

Jshd efficiencyThe .jshd files offer a way to restructure data such that expensive geospatial operations can be avoided during simulation runtime. Even so, practical usage is revealing that the cell grid tends to be higher resolution than input data and the resulting file can be somewhat large. This leads to two different aspects of external data preprocessing that are worth considering further: time to compute and space to store. In general, time may be less immediately actionable than improvements on space.

Time to compute

At a high level, we prefer that expensive compute happens in preprocessing where possible. Even so, the .jshd files really can take some time to build. Part of this has to do with geospatial operations which are unavoidably expensive. To help facilitate this process, [@Nick Gondek](https://schmidt-dse.slack.com/team/U06005HP43T) already enabled parallelism in the preprocessing step. However, another aspect of this has to do with just the mechanical process of writing large amounts of data, even if much of those data are duplicative.

Problem

We re-compute the transformation from Earth space to grid space for each cell per input file per step. This involves a write per cell and often the cell values are duplicative so this entire loop repeats potentially many more times than strictly necessary.

Proposal

1. The transformation could be saved between files, cutting out the geospatial operations if the files being preprocessed are consistent in structure / metadata.
2. We could see if there are benefits from pre-computing an Earth to grid space transformation prior to operations thanks to memory locality... even if it is not re-used across files.

Questions

* After closely inspecting the preprocessing logic, might precomputing the transformation might save time?
* What does JFR say about current bottleneck right now?

Differential

* Given that the precompute step was always expected to be expensive to save time later, is this a good use of our resources or should we focus on space considerations?
* Should the data structure itself change to avoid repetitive writes?

Space to store

The .jshd is a file whose structure is optimized for a fast read into memory using a structure which is fast to leverage during simulation. Almost all of our efforts have been focused on compute over space (for good reason!). Even so, we are encountering some drawbacks on this approach as we get further into practical modeling.

Problem

Josh Cloud's architecture simultaneously preserves privacy, reduces costs, and works to maintain an open source pipeline (no vendor lock in). However, the resulting ephemeral computation approach limits external data size. Therefore, in consideration of our choice of space inefficiency in the .jshd format in exchange for computational speed, this inadvertently limits the size of external data which can run using that platform even if the original external data would have fit into GCP limits. This is because that external data are expanded due to projection into grid space which may be higher resolution than the external data.

Example: The dense flat binary layout of .jshd (header + width × height × timesteps × 8 bytes of raw IEEE 754 doubles) means that a 1000×1000 grid over 30 timesteps already reaches ~240 MB — before base64 encoding for cloud transmission adds another ~33%.

Proposals

Thanks to Apache Commons Compress, open source offers a very low effort Java-native way to compress data while being written from preprocessing and transparently decompress in Josh Cloud when run. Specifically, we propose using the XZ format (LZMA2 codec). Compress exposes XZ directly as a stream compressor (XZCompressorOutputStream / XZCompressorInputStream), which is simpler to integrate and achieves equivalent or better compression without the overhead of an archive structure. This could be a .jsdz format (jshd XZ-compressed) that would dramatically expand what can run in Josh Cloud. The implementation fits naturally into the existing GridSerializationStrategy decorator pattern. We imagine that a CompressedGridSerializationStrategy wraps any inner strategy by intercepting the OutputStream on write and InputStream on read, identical in structure to the existing Base64StringGridSerializationStrategy. Commons Compress also supports magic-byte auto-detection on read, so the decompression path does not need to know the algorithm in advance.

Limitations

Apache Commons Compress is not compatible with TeaVM. Therefore, the .jsdz format would intentionally be scoped to JVM execution paths (local CLI and Josh Cloud) for now. The browser editor would continue to use .jshd only.

Questions

* How much time does XZ cost in write and read?
* How much savings can we expect practically in Josh Cloud?
* Architecturally, how easy will it be to transparently slot Apache Commons Compress in the middle?

Differential

* We could switch to a more optimized data structure, potentially solving for compute at the same time — though this is higher effort and higher risk than wrapping compression around the existing format.
* We may consider waiting until we learn more from Maya and Lucia given upcoming use before we engage in forward development.
* We could leverage GCS to hold large blobs instead of trying to make files smaller, though this risks some degree of vendor lock in.
* A streaming write path in JshdUtil (writing doubles directly to the compressor stream rather than pre-allocating a full ByteBuffer) would reduce peak memory during preprocessing and is a natural long-term companion to this change. However, due to implementation complexity, we currently recommend deferring this from the initial implementation.
* The current serializeToBytes pre-allocates a ByteBuffer of full file size in heap. For large grids this is a secondary concern but worth revisiting if memory pressure becomes an issue in preprocessing.
* What does JFR say about the current preprocessing bottleneck?
    * Based on code inspection the geospatial mapper loop is expected to dominate so the final serialization step should be minor.
    * This should be confirmed before investing in write-path optimizations.

GondekNP:
## Summary

- Introduces `XzGridSerializationStrategy` — an XZ/LZMA2 compression decorator wrapping any `GridSerializationStrategy`, following the same pattern as `Base64StringGridSerializationStrategy`
- Adds `.jshdz` as a compressed output/input format for `preprocess` and `run` commands (JVM only; browser editor continues to use `.jshd`)
- Wires compression transparently: passing a `.jshdz` path to `preprocess` selects the XZ serializer automatically, and `run` now uses `MultiFormatExternalGetter` to handle both extensions
- `inspectJshd` command updated to accept `.jshdz` files

Closes #425

## New files

| File | Role |
|---|---|
| `XzGridSerializationStrategy` | Decorator: wraps inner strategy with XZ compress/decompress |
| `JshdZExternalGetter` | Loads `.jshdz` files in simulations (JVM only) |
| `MultiFormatExternalGetter` | Routes `.jshd`/`.jshdz` by extension; used by `RunCommand` |
| `XzGridSerializationStrategyTest` | Round-trip + compression-ratio tests |
| `JshdZExternalGetterTest` | Load and extension-rejection tests |

## TeaVM safety

`JoshSimFacadeUtil.runSimulation(InputOutputLayer, ...)` (called by the TeaVM path) is unchanged — it still creates `JshdExternalGetter` internally. A new overload accepting `ExternalResourceGetter` is used only by JVM callers. Neither `XzGridSerializationStrategy` nor any XZ import appears in the TeaVM compilation path.

## Test plan

- [x] `./gradlew test` — 1966 tests pass, 0 failures
- [x] `preprocess ... output.jshdz` produces a valid compressed file (~6000× smaller on constant test data)
- [x] `inspectJshd output.jshdz data 0 0 0` returns correct value
- [x] `inspectJshd output.jshdz data --to-csv` exports full grid to CSV correctly
- [x] Values match equivalent `.jshd` output for the same input

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -44,6 +44,10 @@ dependencies {
   // Apache Commons Collections for LRUMap
   implementation("org.apache.commons:commons-collections4:4.5.0")
 
+  // Apache Commons Compress + XZ library for LZMA2 compression (jshdz format, JVM only)
+  implementation("org.apache.commons:commons-compress:1.27.1")
+  implementation("org.tukaani:xz:1.10")
+
   // Apache Commons CSV for CSV handling
   implementation("org.apache.commons:commons-csv:1.10.0")
 
--- a/llms-full.txt
+++ b/llms-full.txt
@@ -65,10 +65,10 @@ java -jar joshsim-fat.jar run simulation.josh Main --data ./data -o results.netc
 ```
 
 #### `preprocess` - Data Preprocessing
-Converts external geospatial data files into optimized binary `.jshd` format for efficient simulation loading.
+Converts external geospatial data files into optimized binary `.jshd` format for efficient simulation loading. Output can also be written as `.jshdz` (XZ/LZMA2-compressed JSHD) by using that extension — **JVM only**, not supported in the browser editor.
 
 ```bash
-java -jar joshsim-fat.jar preprocess <script.josh> <SimulationName> <dataFile> <variable> <units> <output.jshd> [OPTIONS]
+java -jar joshsim-fat.jar preprocess <script.josh> <SimulationName> <dataFile> <variable> <units> <output.jshd|output.jshdz> [OPTIONS]
 ```
 
 **Parameters:**
@@ -77,10 +77,10 @@ java -jar joshsim-fat.jar preprocess <script.josh> <SimulationName> <dataFile> <
 - `<dataFile>`: Path to input data file (NetCDF, GeoTIFF, etc.)
 - `<variable>`: Variable name or band number to extract
 - `<units>`: Units of the data for simulation use
-- `<output.jshd>`: Path for output preprocessed file
+- `<output.jshd>` or `<output.jshdz>`: Path for output preprocessed file. Use `.jshdz` extension to produce an XZ-compressed file (significantly smaller, JVM only).
 
 **Options:**
-- `--amend`: Amend existing JSHD file rather than overwriting
+- `--amend`: Amend existing JSHD/JSHDZ file rather than overwriting
 - `--crs <CRS>`: Coordinate Reference System for reading input file
 - `--x-coord <name>`: Name of X coordinate dimension
 - `--y-coord <name>`: Name of Y coordinate dimension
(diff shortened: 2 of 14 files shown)


sampottinger:
Exciting!

PR #433, opened 2026-04-26
batchRemote: thread --custom-tag and --replicate-start through dispatch

GondekNP:
## Summary

Restores `--custom-tag` parity on `batchRemote` (lost in #423) and adds `--replicate-start` for pool/resume sweep workflows. Both flags also land on `run` and `runRemote` so the same `.josh` file works on every run mode.

After this PR, joshpy ([joshpy#39](https://github.com/SchmidtDSE/joshpy/pull/39)) can:
- Inject `run_hash` (or any custom tag) into `exportFiles` paths without pre-interpolating tokens in `sim.josh` before upload.
- Implement `pool`/`replace`/`skip`/`fail` collision policies for re-dispatch by computing replicate offsets pre-dispatch.

### CLI
- `RunCommand` — adds `--replicate-start` (default `0`), mutex with `--replicate-index`.
- `RunRemoteCommand` — same flag, threaded via `RunRemoteContextBuilder.withReplicateNumber()` (consumed by `RunRemoteLocalLeaderStrategy.createWorkerTasks` for the local-leader path); also sent as a `replicateStart` form field on the offload-leader path.
- `BatchRemoteCommand` — adds `--custom-tag` and `--replicate-start`. `parseCustomTags()` mirrors `RunCommand.parseCustomParameters()` (rejects reserved names `replicate`/`step`/`variable`).

### Dispatch chain
- `RemoteBatchTarget.dispatch()` — extended signature: `(jobId, prefix, simulation, replicates, customTags, replicateStart)`.
- `KubernetesTarget` — emits `JOSH_CUSTOM_TAGS` (newline-delimited `key=value` lines) and `JOSH_REPLICATE_OFFSET` env vars on the Job spec.
- `HttpBatchTarget` — propagates as `customTags` and `replicateStart` form fields.
- Both new env vars / form fields are emitted **only when non-default**, so existing dispatches produce byte-identical Job specs and form bodies.

### Server-side
- `JoshSimBatchHandler` — parses both new form fields, **wires `TemplateStringRenderer` through `JvmInputOutputLayerBuilder` for the first time** on this code path (it was never wired before, so `{run_hash}` and other custom tags silently no-op'd on the HTTP batch path even prior to this PR — fixing this is partly a bug fix), and offsets the replicate loop by `replicateStart`. Validates negative/non-numeric `replicateStart` and reserved/malformed `customTags`.
- `JoshSimLeaderHandler` — accepts the new `replicateStart` form field and offsets `WorkerTask` replicate numbering for the offload-leader path.

### Pod entrypoint
- `cloud-img/run-entrypoint.sh` — reads `JOSH_CUSTOM_TAGS` via POSIX `while IFS= read -r line` (no `jq` dep — keeps `Dockerfile.batch` JRE-only), emits one `--custom-tag=key=value` flag per non-empty line. Computes `REPLICATE_INDEX = JOB_COMPLETION_INDEX + JOSH_REPLICATE_OFFSET`. `preprocess-entrypoint.sh` is unchanged — preprocessing is tag-free by design.

## Encoding decisions

- **Newline-delimited `key=value` for tags**: matches `RunCommand`'s existing `String[] customTags` shape exactly, parseable in pure POSIX shell, no new image deps. Newlines in tag values are not supported (joshpy's typical tag values are alphanumeric in practice).
- **Defaults are no-ops**: empty tag map and `replicateStart=0` produce byte-identical specs to before this PR.

### Unit (~12 new)
- `RunCommandReplicatesTest` — `--replicate-start=5 --replicates=2` produces CSV rows with replicate column = 5 and 6, none = 0; mutex with `--replicate-index`; rejects negative.
- `BatchRemoteCommandTest` — schema (`--custom-tag` is `String[]`, `--replicate-start` is `int` with default `"0"`); `parseCustomTags` rejects reserved names + malformed entries.
- `KubernetesTargetTest` — env vars omitted at default; newline-joined value when tags present; offset emitted iff non-zero.
- `HttpBatchTargetTest` — form fields omitted at default; URL-encoded round-trip of newlines verified by capturing the outgoing `HttpRequest` body.
- `JoshSimBatchHandlerTest` — rejects malformed `customTags`, reserved-name `customTags`, negative `replicateStart`, non-numeric `replicateStart`.
- Existing `BatchJobStrategyTest`, `HttpBatchTargetTest`, `KubernetesTargetTest` updated to the new signature with `Map.of(), 0`.

### Integration (`.github/workflows/test-k8s.yaml`)

New step inside the existing Kind CI workflow dispatches:

```sh
java -jar build/libs/joshsim-fat.jar batchRemote K8sCustomTagTest \
  --target=ci-k8s \
  --minio-prefix=batch-jobs/k8s-customtag/inputs/ \
  --require-prestaged \
  --replicates=2 \
  --replicate-start=2 \
  --custom-tag=run_hash=ci-test
```

The `.josh` script templates both `{run_hash}` and `{replicate}` into the export path:
```
minio://josh-test-bucket/k8s-customtag-results/{run_hash}/replicate-{replicate}.csv
```

The verification step asserts:
- `ci-test/` directory exists under `k8s-customtag-results/` (proves `{run_hash}` flows through K8s env var → entrypoint → `--custom-tag` flag → `TemplateStringRenderer`)
- `replicate-2.csv` and `replicate-3.csv` are present and non-empty (proves the offset reaches the pod via `JOSH_REPLICATE_OFFSET + JOB_COMPLETION_INDEX`)
- `replicate-0.csv` is absent (proves the pods didn't fall back to `JOB_COMPLETION_INDEX` alone)

All three assertions would fail before this PR.

### Local smoke
`joshsim run --replicates=2 --replicate-start=5 --custom-tag=run_hash=local-test` produces exactly `local-test/replicate-5.csv` and `local-test/replicate-6.csv` (no `replicate-0.csv`) — same logic the pod entrypoint executes.

## Test plan
- [x] \`./gradlew test checkstyleMain checkstyleTest\` — all pass
- [x] \`./gradlew fatJar\` — builds clean
- [x] \`--help\` on \`run\`, \`runRemote\`, \`batchRemote\` shows the new flags
- [x] Local end-to-end smoke via \`run\` — output paths verified
- [ ] CI: existing K8s integration tests still pass (verifies backward-compat)
- [ ] CI: new K8s integration step verifies the new flags end-to-end against Kind

### DIFF
--- a/src/main/java/org/joshsim/pipeline/target/KubernetesTarget.java
+++ b/src/main/java/org/joshsim/pipeline/target/KubernetesTarget.java
@@ -91,7 +91,9 @@ public class KubernetesTarget implements RemoteBatchTarget {
       String jobId,
       String minioPrefix,
       String simulation,
-      int replicates
+      int replicates,
+      Map<String, String> customTags,
+      int replicateStart
   ) throws Exception {
     String secretName = SECRET_NAME_PREFIX + jobId;
 
@@ -128,7 +130,8 @@ public class KubernetesTarget implements RemoteBatchTarget {
                         )
                         .withEnv(buildEnvVars(
                             secretName, jobId,
-                            minioPrefix, simulation
+                            minioPrefix, simulation,
+                            customTags, replicateStart
                         ))
                         .withCommand(
                             ENTRYPOINT,
@@ -209,7 +212,9 @@ public class KubernetesTarget implements RemoteBatchTarget {
       String secretName,
       String jobId,
       String minioPrefix,
-      String simulation
+      String simulation,
+      Map<String, String> customTags,
+      int replicateStart
   ) {
     List<EnvVar> envVars = new ArrayList<>();
     envVars.add(secretEnvVar(
@@ -227,7 +232,38 @@ public class KubernetesTarget implements RemoteBatchTarget {
     envVars.add(plainEnvVar("JOSH_JOB_ID", jobId));
     envVars.add(plainEnvVar("JOSH_MINIO_PREFIX", minioPrefix));
     envVars.add(plainEnvVar("JOSH_SIMULATION", simulation));
+    if (customTags != null && !customTags.isEmpty()) {
+      envVars.add(plainEnvVar(
+          "JOSH_CUSTOM_TAGS", encodeCustomTags(customTags)
+      ));
+    }
+    if (replicateStart != 0) {
+      envVars.add(plainEnvVar(
+          "JOSH_REPLICATE_OFFSET", String.valueOf(replicateStart)
+      ));
+    }
     return envVars;
+  }
+
+  /**
+   * Encodes custom tags as a newline-delimited {@code key=value} string for the pod entrypoint.
+   *
+   * <p>The pod-side {@code run-entrypoint.sh} reads this env var and emits one
+   * {@code --custom-tag=key=value} flag per line. Newline delimiting matches
+   * {@code RunCommand}'s {@code String[] customTags} shape exactly and avoids
+   * needing a JSON parser ({@code jq}) in the JRE-only batch image.</p>
+   */
+  private static String encodeCustomTags(Map<String, String> customTags) {
+    StringBuilder sb = new StringBuilder();
+    boolean first = true;
+    for (Map.Entry<String, String> entry : customTags.entrySet()) {
+      if (!first) {
+        sb.append('\n');
+      }
+      first = false;
+      sb.append(entry.getKey()).append('=').append(entry.getValue());
+    }
+    return sb.toString();
   }
 
   private static EnvVar secretEnvVar(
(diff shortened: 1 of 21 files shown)


src/main/java/org/joshsim/pipeline/target/KubernetesTarget.java
```diff
@@ -227,9 +232,40 @@ private List<EnvVar> buildEnvVars(
...
+  /**
+   * Encodes custom tags as a newline-delimited {@code key=value} string for the pod entrypoint.
+   *
+   * <p>The pod-side {@code run-entrypoint.sh} reads this env var and emits one
+   * {@code --custom-tag=key=value} flag per line. Newline delimiting matches
+   * {@code RunCommand}'s {@code String[] customTags} shape exactly and avoids
+   * needing a JSON parser ({@code jq}) in the JRE-only batch image.</p>
+   */
+  private static String encodeCustomTags(Map<String, String> customTags) {
+    StringBuilder sb = new StringBuilder();
```
GondekNP: This is not very DRY with httpTarget - prob just factor both out as a helper maybe - BatchArgUtil.encodeCustomTags()

GondekNP: Done in c01b0c6c — extracted to `BatchArgUtil.encodeCustomTags` (your suggested name) in the same package. Both targets now call it; helper is covered by `BatchArgUtilTest` (5 cases including LinkedHashMap order preservation and values containing `=`). Switched the implementation to `Collectors.joining` while moving it.

PR #439, opened 2026-05-14
Update llms-full.txt to describe latest dev.

sampottinger:
Update llms-full.txt to describe the latest dev version as of May 14, 2026. Expanded to include batch run commands.


llms-full.txt
```diff
@@ -169,113 +205,36 @@ java -jar joshsim-fat.jar inspect-exports <script.josh> <SimulationName> [OPTION
...
 **Options:**
-- `--port <port>`: Server port (default: 8080)
+- `--concurrent-workers <n>`: Number of concurrent workers allowed (default: 1)
 - `--host <host>`: Server host (default: localhost)
+- `--port <port>`: Server port (default: 8080)
+- `--serial-patches`: Run patches in serial instead of parallel
+- `--use-http2`: Enable HTTP/2 support (streaming)
+- `--worker-url <url>`: URL for worker requests (default: http://0.0.0.0:8085/runReplicate)
 
-**Server Endpoints:**
```
sampottinger, line 185: I kinda figured we weren't suggesting users hit the server endpoints directly but LMK what you think @GondekNP. Can also reference josh py from here.

GondekNP: I agree with the approach to maybe minimize it it llms-full.txt.... In developing over at `joshpy`, it became helpful to have the endpoints documented clearly in this way for the CLI wrapper to know what `josh` was doing, but I agree I don't think this is something a 'real' user would have any need for and the llms would likely just be confused by it

PR #440, opened 2026-05-26
Add MCP server (Phase 1): stdio transport, local backend, four tools

GondekNP:
## Summary

Implements Phase 1 of the Josh MCP (Model Context Protocol) server per
[`IMPLEMENTATION_PLAN.md`](./IMPLEMENTATION_PLAN.md). Adds a new `mcp`
picocli subcommand that starts a stdio JSON-RPC server exposing four
Josh operations as MCP tools, enabling local LLM clients (opencode,
Claude Desktop, etc.) to drive Josh against files on the user's disk.

```
java -jar joshsim-fat.jar mcp
```

opencode config:
```json
{
  "mcp": {
    "josh": {
      "type": "local",
      "command": ["java", "-jar", "/path/to/joshsim-fat.jar", "mcp"]
    }
  }
}
```

## Tools exposed

| Tool                  | Wraps                                       |
|-----------------------|---------------------------------------------|
| `validate_simulation` | `JoshSimCommander.getJoshProgram`           |
| `discover_config`     | `JoshConfigDiscoveryVisitor` + formatter    |
| `preprocess_data`     | `PreprocessUtil.preprocess`                 |
| `run_simulation`      | `JoshSimFacadeUtil.runSimulation`           |

`run_simulation` exposes a minimal v1 parameter set: `script`,
`simulation`, `replicates`, `serialPatches`, `seed`. Additional knobs
from `RunCommand` (`--data`, `--custom-tag`, `--output-steps`, etc.)
are deferred to a follow-up.

## Architecture

Per the plan, the v1 layout is chosen to make Phase 2 (remote compute
backend) and the eventual hosted mode additive rather than a rewrite:

- `mcp/Backend.java` — interface with `Path`-typed args; one
  implementation in this PR: `LocalBackend` (in-JVM via existing
  facades). Future `RemoteBackend` (Josh Cloud / k8s) slots in
  unchanged at the tool layer.
- `mcp/tool/local/` — path-based tools for v1. Sibling
  `mcp/tool/hosted/` package planned for resource-based tools when
  hosted mode lands.
- `mcp/JoshPaths.resolve(String)` — single seam for path resolution.
  Today does `Paths.get(arg).toAbsolutePath().normalize()`; future
  sandboxing / root-fencing plugs in here.
- `mcp/StderrOutputOptions` — routes all Josh diagnostic output to
  `System.err`. Stdio MCP corrupts on any stdout writes.

## Stdout discipline

The single most important correctness invariant for stdio MCP: only
JSON-RPC framing on `System.out`. Achieved by:

1. `StderrOutputOptions` overriding `OutputOptions.printInfo` to write
   to stderr.
2. `LocalBackend.discoverConfig` invoking the discovery logic
   directly rather than calling `DiscoverConfigCommand.call()` (which
   writes results to `System.out` — `DiscoverConfigCommand.java:87,89`).
3. `McpStdoutCleanlinessTest` proves the invariant by buffering
   `System.out` during backend operations and asserting it stays empty.

`McpCommand` uses `Thread.currentThread().join()` to keep the main
thread alive — same idiom as the existing `ServerCommand`. (An
earlier draft read `System.in` from the main thread, which would have
raced with the SDK's own stdin reader; fixed before merging.)

## Tests

22 tests across 3 classes, all passing:
- `JoshPathsTest` (6) — path resolution.
- `LocalBackendTest` (11) — validate, discoverConfig, result POJOs,
  output routing.
- `McpStdoutCleanlinessTest` (5) — buffer-and-assert that the stdout
  stream is untouched by backend operations.

End-to-end `run_simulation` and a full stdio integration test against
the MCP Inspector are deliberately out of scope for this PR and will
follow.

## Verification

Locally:
- `./gradlew compileJava` — clean
- `./gradlew checkstyleMain` — clean (zero violations as required)
- `./gradlew test --tests "org.joshsim.mcp.*"` — 22/22 pass
- `./gradlew fatJar` — succeeds (~76 MB; bundles `io.modelcontextprotocol.sdk:mcp:1.1.3`)
- Smoke test: a JSON-RPC `initialize` request to
  `java -jar build/libs/joshsim-fat.jar mcp` produces a well-formed
  response on stdout.

## Test plan

- [ ] CI builds clean (compile, checkstyle, full test suite).
- [ ] Manual: drive the server from opencode with a small example
      simulation from `examples/`, exercising each of the four tools.
- [ ] Manual: run the MCP Inspector
      (`npx @modelcontextprotocol/inspector java -jar build/libs/joshsim-fat.jar mcp`)
      to verify the tools list and schemas render correctly.

## Out of scope (follow-ups)

Per the plan:
- **Phase 2:** `RemoteBackend` for Josh Cloud / k8s compute, plus the
  `execution` arg on `preprocess_data` / `run_simulation`.
- **Phase 3:** `inspect_exports`, `inspect_jshd`, and MCP resources
  exposing `.jshd` / `.jshc` files.
- Extracting `RunPipeline` / `RemoteRunPipeline` from the existing CLI
  command classes (the plan calls for this; Phase 1 ships without it
  to keep this PR scoped).
- Expanding `run_simulation`'s parameter surface to full
  `RunCommand` parity.

https://claude.ai/code/session_01WnoBnDFGm6mmo7Rsn4ua7v

---
_Generated by [Claude Code](https://claude.ai/code/session_01WnoBnDFGm6mmo7Rsn4ua7v)_

### DIFF
--- a/examples/test_mcp.sh
+++ b/examples/test_mcp.sh
@@ -38,30 +38,13 @@ for tool in validate_simulation discover_config preprocess_data run_simulation; 
 done
 echo "PASS: all four tools advertised"
 
-# Prepare a tiny, self-contained simulation in a temp dir.
+# Use an existing self-contained example (no external data) for validate + run.
 WORK="$(mktemp -d)"
 trap 'rm -rf "$WORK"' EXIT
-RUN_JOSH="$WORK/run.josh"
-RUN_OUT="$WORK/run_out.csv"
-cat > "$RUN_JOSH" <<JOSH
-start simulation TestSim
-  grid.size = 100 m
-  grid.low = 0 degrees latitude, 0 degrees longitude
-  grid.high = 0.1 degrees latitude, 0.1 degrees longitude
-  grid.patch = "Default"
-  steps.low = 0 count
-  steps.high = 2 count
-  exportFiles.patch = "file://$RUN_OUT"
-end simulation
-start patch Default
-  Tree.init = create 2 count of Tree
-  export.treeCount.step = count(Tree)
-end patch
-start organism Tree
-  age.init = 0 count
-  age.step = prior.age + 1 count
-end organism
-JOSH
+RUN_JOSH="examples/simulations/simple.josh"
+RUN_SIM="TestSimpleSimulation"
+RUN_OUT="/tmp/simple_josh.csv"  # export path declared inside simple.josh
+rm -f "$RUN_OUT"
 
 # 2. validate_simulation — should report a successful validation.
 echo "--- validate_simulation ---"
@@ -74,7 +57,7 @@ echo "PASS: validate_simulation"
 # 3. run_simulation — should complete and write the export CSV.
 echo "--- run_simulation ---"
 RUN=$(inspect --method tools/call --tool-name run_simulation \
-  --tool-arg script="$RUN_JOSH" --tool-arg simulation=TestSim)
+  --tool-arg script="$RUN_JOSH" --tool-arg simulation="$RUN_SIM")
 echo "$RUN"
 echo "$RUN" | grep -qi "completed" || { echo "FAIL: run_simulation did not complete"; exit 1; }
 [ -s "$RUN_OUT" ] || { echo "FAIL: run_simulation produced no CSV output"; exit 1; }
--- a/examples/test_mcp.sh
+++ b/examples/test_mcp.sh
@@ -1,0 +1,99 @@
+#!/bin/bash
+# MCP server regression smoke test, driven by the official MCP Inspector CLI.
+#
+# Exercises the stdio MCP server end-to-end the way a real MCP client would:
+#   1. tools/list                — all four tools are advertised
+#   2. validate_simulation       — cheap parse/validate path
+#   3. run_simulation            — local run path (writes a CSV)
+#   4. preprocess_data           — local preprocess path (writes a .jshd)
+#
+# Requires `node`/`npx` on PATH (preinstalled on GitHub-hosted ubuntu runners)
+# and a built fat jar at build/libs/joshsim-fat.jar.
+set -euo pipefail
+
+JAR="build/libs/joshsim-fat.jar"
+# Pin a version here (e.g. @modelcontextprotocol/inspector@0.x.y) once a known-good
+# release is confirmed, so the inspector's own CLI changes can't silently break CI.
+INSPECTOR="@modelcontextprotocol/inspector"
+
+if [ ! -f "$JAR" ]; then
+  echo "ERROR: $JAR not found. Build it with: ./gradlew fatJar"
+  exit 1
+fi
+
+echo "node: $(node --version), npm: $(npm --version)"
+echo "=== MCP Inspector CLI smoke test ==="
+
+# Helper: run the inspector against the stdio server with the given inspector args.
+inspect() {
+  npx --yes "$INSPECTOR" --cli java -jar "$JAR" mcp "$@"
+}
+
+# 1. tools/list — every tool must be advertised.
+echo "--- tools/list ---"
+LIST=$(inspect --method tools/list)
+echo "$LIST"
+for tool in validate_simulation discover_config preprocess_data run_simulation; do
+  echo "$LIST" | grep -q "\"$tool\"" || { echo "FAIL: tool '$tool' not advertised"; exit 1; }
+done
+echo "PASS: all four tools advertised"
+
+# Prepare a tiny, self-contained simulation in a temp dir.
+WORK="$(mktemp -d)"
+trap 'rm -rf "$WORK"' EXIT
+RUN_JOSH="$WORK/run.josh"
+RUN_OUT="$WORK/run_out.csv"
+cat > "$RUN_JOSH" <<JOSH
+start simulation TestSim
+  grid.size = 100 m
+  grid.low = 0 degrees latitude, 0 degrees longitude
+  grid.high = 0.1 degrees latitude, 0.1 degrees longitude
+  grid.patch = "Default"
+  steps.low = 0 count
+  steps.high = 2 count
+  exportFiles.patch = "file://$RUN_OUT"
+end simulation
+start patch Default
+  Tree.init = create 2 count of Tree
+  export.treeCount.step = count(Tree)
+end patch
+start organism Tree
+  age.init = 0 count
+  age.step = prior.age + 1 count
+end organism
+JOSH
+
+# 2. validate_simulation — should report a successful validation.
+echo "--- validate_simulation ---"
+VALIDATE=$(inspect --method tools/call --tool-name validate_simulation \
+  --tool-arg script="$RUN_JOSH")
+echo "$VALIDATE"
+echo "$VALIDATE" | grep -qi "Validated Josh script" || { echo "FAIL: validate_simulation"; exit 1; }
+echo "PASS: validate_simulation"
+
+# 3. run_simulation — should complete and write the export CSV.
+echo "--- run_simulation ---"
+RUN=$(inspect --method tools/call --tool-name run_simulation \
+  --tool-arg script="$RUN_JOSH" --tool-arg simulation=TestSim)
+echo "$RUN"
+echo "$RUN" | grep -qi "completed" || { echo "FAIL: run_simulation did not complete"; exit 1; }
+[ -s "$RUN_OUT" ] || { echo "FAIL: run_simulation produced no CSV output"; exit 1; }
+echo "PASS: run_simulation (CSV written)"
+
+# 4. preprocess_data — convert the committed test GeoTIFF into a .jshd.
+echo "--- preprocess_data ---"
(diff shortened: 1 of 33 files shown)


src/main/java/org/joshsim/mcp/tool/local/PreprocessTool.java
```diff
@@ -0,0 +1,154 @@
...
+ * Registers the {@code preprocess_data} MCP tool.
+ *
+ * <p>Preprocessing converts an external geospatial dataset (NetCDF, GeoTIFF, or existing
+ * {@code .jshd} file) into Josh's binary {@code .jshd} format, aligned to the simulation grid
+ * defined in a {@code .josh} script. The resulting file can then be referenced from the script
+ * using external data expressions.</p>
+ */
+public final class PreprocessTool {
+
+  private static final String INPUT_SCHEMA = "{"
```
GondekNP: Let's factor this out to an asset somewhere rather than creating this JSON inline

src/main/java/org/joshsim/mcp/tool/local/PreprocessTool.java
```diff
@@ -0,0 +1,154 @@
...
+          Path outputFile = JoshPaths.resolve(outputArg);
+
+          Backend.PreprocessResult result = backend.preprocess(
+              script, simArg, dataFile, variableArg, unitsArg, outputFile, Optional.empty()
+          );
+          return CallToolResult.builder()
+              .addTextContent(result.getMessage())
+              .isError(!result.isSuccess())
+              .build();
+        })
```
GondekNP: Here and for all of the builder methods - let's factor out this handling / validation stuff into a method and call it via the builder. Each of these should basically look the same, with the  handling / validation looking different between implementations. 

Currently this huge anonymous method is confusing

src/main/java/org/joshsim/mcp/tool/local/RunSimulationTool.java
```diff
@@ -0,0 +1,149 @@
...
+            + "Josh simulations define a spatial grid, organism types, and step logic. "
+            + "Each step corresponds to one time unit (typically a year). "
+            + "Run validate_simulation first to catch any script errors before a long simulation."
+        )
+        .inputSchema(jsonMapper, INPUT_SCHEMA)
+        .build();
+
+    SyncToolSpecification spec = SyncToolSpecification.builder()
+        .tool(tool)
+        .callHandler((exchange, request) -> {
```
GondekNP: see above comment

src/main/java/org/joshsim/mcp/LocalBackend.java
```diff
@@ -0,0 +1,261 @@
...
+          "Successfully preprocessed data to " + outputFile);
+    } catch (IllegalArgumentException e) {
+      return new PreprocessResult(false, e.getMessage());
+    } catch (Exception e) {
+      return new PreprocessResult(false, "Preprocessing failed: " + e.getMessage());
+    }
+  }
+
+  @Override
+  public RunSimulationResult runSimulation(
```
GondekNP, line 157: Now that we have several places we are calling the actual guts of `run`, we need to factor out the basic logic into a `runUtil` just like the above `preprocessUtil`. These MCP implementations should be very, very thin and shouldn't touch application logic AT ALL.

GondekNP:
Pushed `a78c9734` addressing the review feedback.

**Comment 1 — inline JSON schemas → assets.** Each tool's `INPUT_SCHEMA` now lives in a dedicated resource file under `src/main/resources/org/joshsim/mcp/tool/local/` (`validate_simulation.schema.json`, `discover_config.schema.json`, `preprocess_data.schema.json`, `run_simulation.schema.json`), loaded via a new `ToolSchemas.load(toolName)` classpath helper. Confirmed they're packaged into the fat jar and render correctly over a `tools/list` request.

**Comments 2 & 3 — huge anonymous handler methods.** Each tool's call-handler body is lifted out of the lambda into a named `private static handle(request, backend)` method, so every `register()` is now uniform SDK scaffolding and the tool-specific logic reads top-to-bottom. Added `ToolHandlers` with a shared `errorResult(...)` (was duplicated) and a `requireString(...)` helper that collapses the repetitive null/blank argument checks.

Also added `ToolSchemasTest`. `compileJava` / `checkstyleMain` clean; 24 MCP tests pass.

**Comment 4 — `RunUtil` extraction.** This is the bigger DRY refactor (extract a `RunUtil` à la `PreprocessUtil` and migrate **both** `RunCommand` and `LocalBackend` onto it, so the MCP server doesn't reimplement run logic). To keep that change clean and reviewable on its own, I'm doing it as a **separate PR on a dedicated feature branch** rather than piling it into this one. It includes a deep-dive on which `RunCommand` params the MCP surface should expose vs. default vs. hardcode — the guiding rule being that the MCP server must not behave differently from the jar unless a default is hardcoded and clearly justified (e.g. MinIO creds read from env, never tool args). Will link it here once it's up.

GondekNP:
Follow-up for review comment 4 (the `RunUtil` DRY extraction) is up as #441, stacked on this branch: https://github.com/SchmidtDSE/josh/pull/441

It removes the divergence where the MCP run path skipped Earth-space grid handling, progress metadata, csv precision, and MinIO — `LocalBackend.runSimulation` and `RunCommand` now share one `RunUtil.run` pipeline.

examples/test_mcp.sh
```diff
@@ -0,0 +1,99 @@
...
+done
+echo "PASS: all four tools advertised"
+
+# Prepare a tiny, self-contained simulation in a temp dir.
+WORK="$(mktemp -d)"
+trap 'rm -rf "$WORK"' EXIT
+RUN_JOSH="$WORK/run.josh"
+RUN_OUT="$WORK/run_out.csv"
+cat > "$RUN_JOSH" <<JOSH
+start simulation TestSim
```
GondekNP: Just use one of our existing examples, no need to inline this

src/main/java/org/joshsim/command/McpCommand.java
```diff
@@ -0,0 +1,71 @@
...
+  @Override
+  public Integer call() {
+    StderrOutputOptions output = new StderrOutputOptions();
+    LocalBackend backend = new LocalBackend(output);
+    JoshMcpServer mcpServer = new JoshMcpServer(backend);
+
+    try {
+      mcpServer.start();
+
+      // Block the main thread forever. The SDK owns stdin/stdout on its own threads;
```
GondekNP: Is this best practice for this kind of application in java? I don't hate this but wonder if there's a better solution

src/main/java/org/joshsim/command/RunUtil.java
```diff
@@ -0,0 +1,738 @@
...
+    );
+
+    JoshSimFacadeUtil.runSimulation(
+        valueFactory,
+        geometryFactory,
+        inputOutputLayer,
+        externalGetter,
+        program,
+        simulation,
+        (step) -> {
```
GondekNP: Extract as a helper method, don't inline

GondekNP: Done in b167676a — extracted into a `stepCallback(progressCalculator, output, lastStep)` helper returning a `SimulationStepCallback`; the `runSimulation` call site now just passes `stepCallback(...)`.

examples/test_mcp.sh
GondekNP: Done in b167676a — validate/run now drive `examples/simulations/simple.josh` (`TestSimpleSimulation`, self-contained, no external data) instead of the inlined heredoc. Verified it still completes and writes its export CSV.

src/main/java/org/joshsim/command/McpCommand.java
GondekNP: Good question. The `Thread.currentThread().join()` was copied from `ServerCommand` (it parks the main thread because a thread can never join itself), but it reads as a trick and — worse — the `finally { mcpServer.close(); }` after it never actually ran when the MCP client ends the session with a SIGTERM (join only returns on interrupt, which doesn't happen on signal).

Changed in b167676a to the conventional pattern: park the main thread on a `CountDownLatch` and register a JVM shutdown hook that closes the server and releases the latch. Same "run in foreground until the client stops us" behavior, clearer intent, and cleanup now runs on signal-driven shutdown too (satisfies the "cleanup in finally, never rely on JVM exit" rule in IMPLEMENTATION_PLAN). The truly-clean alternative — exit on stdin EOF rather than waiting for a signal — would need the MCP SDK's stdio transport to expose a close/await hook, which it doesn't here; real clients signal the subprocess anyway. Left `ServerCommand` as-is since it's outside this PR's scope.

GondekNP: Decision: keeping it consistent with `ServerCommand` for now. Reverted to the same `Thread.currentThread().join()` idiom (3d794716) and left a NOTE marking the shared smell, so both commands can be revisited together later rather than having one drift ahead. The latch + shutdown-hook approach is the likely cleanup when we do.

PR #441, opened 2026-05-26
DRY: extract RunUtil so CLI and MCP share one run pipeline

GondekNP:
Stacked on the Phase 1 MCP branch (`claude/eloquent-dirac-hmmCi`, PR #440). Targets that branch rather than `dev` because it modifies `LocalBackend`, which only exists on the Phase 1 branch. Merge after / together with #440.

## Why

PR #440 review [comment 4](https://github.com/SchmidtDSE/josh/pull/440#discussion_r3305837284): *"factor out the basic logic into a `runUtil` just like the above `preprocessUtil`. These MCP implementations should be very, very thin and shouldn't touch application logic AT ALL."*

The MCP `LocalBackend.runSimulation` had reimplemented the run loop, duplicating `RunCommand` — and **diverging** from it. The standalone MCP loop skipped Earth-space grid setup, progress metadata, csv precision, and MinIO. Per the guiding rule that *the MCP server should not behave differently from the jar unless a default is hardcoded and clearly justified*, both now drive the same pipeline.

## What

- **New `command/RunUtil.java`** — the single in-JVM run pipeline. `RunUtil.run(RunOptions, OutputOptions)` holds seed setup, grid-search job expansion, program init, Earth-space detection, and the per-replicate loop. `RunOptions` is an immutable builder whose defaults match the `run` CLI command's option defaults. `RunResult` carries success/last-step plus a typed init-failure step and a simulation-not-found flag so each caller derives its own outcome (exit code vs MCP result).
- **`RunCommand.call()` → thin shim** — validate flags, build `RunOptions`, call `RunUtil.run`, map `RunResult` to an exit code. No CLI behavior change.
- **`LocalBackend.runSimulation` → thin shim** — builds the five MCP-exposed knobs (`script`, `simulation`, `replicates`, `serialPatches`, `seed`); everything else falls through to CLI defaults.

### Hardcoded MCP defaults (documented inline in `LocalBackend`)

| Knob | MCP behavior | Why |
|------|--------------|-----|
| `enableProfiler` | off | developer feature, no MCP surface in v1 |
| `MinioOptions` | default instance | `MinioOptions` already resolves CLI flags → config file → **environment**, so models that export to MinIO work over MCP, with creds coming from env rather than tool arguments (keeps secrets out of LLM chat history) |

Every other knob (`replicateStart`, `csvPrecision`, `exportQueueSize`, `useFloat64`, `outputSteps`, `dataFiles`, `customParameters`) uses the same default as the CLI. Phase-2 MCP params (`dataFiles`/`customParameters`/`outputSteps`) are already supported by `RunOptions`; only their MCP tool-schema exposure is deferred.

## Tests & verification

- New `RunUtilTest`: real end-to-end run (asserts CSV written), simulation-not-found, program-init failure, seed determinism. `runSimulation` is now covered end-to-end (was a known gap).
- Existing `RunCommand*` tests pass unchanged — CLI regression net.
- `./gradlew test checkstyleMain checkstyleTest` — clean.
- Manual: `josh run min.josh TestSim --replicates 2` and the MCP `run_simulation` tool both produce identical CSV output through `RunUtil`; the MCP run's diagnostics route to stderr (stdout stays JSON-RPC-only) and the tool returns `Simulation 'TestSim' completed: 1 replicate(s), last step 3`.

### DIFF
--- a/src/main/java/org/joshsim/command/RunUtil.java
+++ b/src/main/java/org/joshsim/command/RunUtil.java
@@ -1,0 +1,665 @@
+/**
+ * Shared simulation-run logic extracted from {@link RunCommand}.
+ *
+ * <p>Holds the full "run a simulation" pipeline — random-seed setup, grid-search job expansion,
+ * program initialization, Earth-space detection, and the per-replicate execution loop — so that
+ * both the {@code run} CLI command and the MCP {@code run_simulation} tool drive identical logic
+ * rather than maintaining divergent reimplementations.</p>
+ *
+ * @license BSD-3-Clause
+ */
+
+package org.joshsim.command;
+
+import java.io.File;
+import java.math.BigDecimal;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import org.joshsim.JoshSimCommander;
+import org.joshsim.JoshSimFacadeUtil;
+import org.joshsim.compat.CompatibilityLayerKeeper;
+import org.joshsim.compat.JvmCompatibilityLayer;
+import org.joshsim.engine.entity.base.MutableEntity;
+import org.joshsim.engine.geometry.EngineGeometryFactory;
+import org.joshsim.engine.geometry.ExtentsUtil;
+import org.joshsim.engine.geometry.PatchBuilderExtentsBuilder;
+import org.joshsim.engine.geometry.grid.GridGeometryFactory;
+import org.joshsim.engine.value.converter.Units;
+import org.joshsim.engine.value.engine.ValueSupportFactory;
+import org.joshsim.engine.value.type.EngineValue;
+import org.joshsim.lang.bridge.GridInfoExtractor;
+import org.joshsim.lang.bridge.ShadowingEntity;
+import org.joshsim.lang.interpret.JoshProgram;
+import org.joshsim.lang.interpret.RecursiveValueResolverFactory;
+import org.joshsim.lang.interpret.TimedRecursiveValueResolverFactory;
+import org.joshsim.lang.interpret.ValueResolverFactory;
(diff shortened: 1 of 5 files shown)


src/main/java/org/joshsim/command/RunUtil.java
```diff
@@ -0,0 +1,665 @@
...
+
+    JoshProgram program = initResult.getProgram().orElseThrow();
+    if (!program.getSimulations().hasPrototype(opts.simulation)) {
+      output.printError("Could not find simulation: " + opts.simulation);
+      return RunResult.simulationNotFound(opts.simulation);
+    }
+
+    // Extract simulation metadata for progress tracking.
+    SimulationMetadata metadata;
+    try {
```
GondekNP, line 550: Let's factor some of these major semantic steps out to be more readable - this is a massive mehtod

GondekNP: Done in 7f6c5f45. Factored `runInternal`'s phases into named private methods so the top-level method now reads as a sequence of steps:

- `buildJobs` / `reportPlannedRun` — grid-search expansion + reporting
- `createValueFactory` / `createInputStrategy` — construction helpers
- `initializeProgram` — first-job init layer + parse/validate
- `configureCompatibilityLayer`, `createProgressCalculator`
- `extractSpatialInfo` — Earth-space detection, returns a `GridSpatialInfo` record
- `executeJobs` — the job/replicate loop, returns an `ExecutionSummary` record
- `reportRunCompletion` — final summary

`runInternal` is now ~30 lines. I also split `runReplicate`'s layer construction into `buildReplicateLayer`, which removed the duplicated builder chain across the Earth-space / non-Earth-space branches and shrank the parameter list. No behavior change — `RunUtilTest`, the `RunCommand*` CLI regression tests, and the MCP tests all pass unchanged.

PR #442, opened 2026-05-27
Fail fast on unsupported grid.size units in preprocess

GondekNP:
## Problem

`josh preprocess` hangs for the full timeout (observed: >600s) when a simulation uses a non-meter spatial unit for `grid.size`, e.g. `grid.size = 16 km`.

Root cause is an **ordering bug**. In `PreprocessUtil.preprocess`, the `unitsSupported()` guard (which only accepts `m`/`meter`/`meters`) ran *after* `gridFactory.build()`. With `km`, the bare numeric value (`16`) reaches `EarthPatchBuilder`, which interprets `getCellSizeMeters()` literally as meters. So `16 km` is treated as `16 m` — a grid ~1000× too fine per axis, i.e. tens of millions of patches allocated in a nested loop before the guard ever runs. The build hangs (or OOMs) instead of producing the clean "Unsupported units" error that already existed a few lines below.

`validate` passes because validation never builds the grid; the `run` path doesn't hang because it skips the Earth-space branch when the size isn't in meters (a separate silent-fallback quirk, not addressed here).

## Fix

Move the existing `unitsSupported()` check ahead of `gridFactory.build()` so an unsupported unit fails immediately, and make the message actionable (point the user at meters). No unit-system refactor — `km` is still not auto-converted; it now produces an instant, clear error instead of a hang.

## Test

Adds `PreprocessUtilGridSizeUnitsTest`: a `grid.size = 16 km` script must throw `IllegalArgumentException` (naming the offending unit) within a bounded time, guarding against the hang regressing.

### DIFF
--- a/src/main/java/org/joshsim/command/PreprocessUtil.java
+++ b/src/main/java/org/joshsim/command/PreprocessUtil.java
@@ -235,6 +235,16 @@ public class PreprocessUtil {
     String endStr = extractor.getEndStr();
     EngineValue size = extractor.getSize();
 
+    // Validate grid size units before building the grid. An unsupported unit such as "km"
+    // would otherwise be silently treated as meters by the Earth-space patch builder,
+    // producing a grid orders of magnitude too fine and hanging preprocessing. Fail fast.
+    if (!unitsSupported(size.getUnits().toString())) {
+      throw new IllegalArgumentException(
+          "Unsupported units for grid size: " + size.getUnits()
+          + ". Grid size must be specified in meters (m); for example use \"1000 m\" "
+          + "instead of \"1 km\".");
+    }
+
     // Build bridge
     CoordinateReferenceSystem crs;
     try {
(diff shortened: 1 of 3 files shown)


src/main/java/org/joshsim/command/PreprocessUtil.java
```diff
@@ -235,6 +235,16 @@ public static void preprocess(File scriptFile, String simulation, String dataFil
     String endStr = extractor.getEndStr();
     EngineValue size = extractor.getSize();
 
+    // Validate grid size units before building the grid. An unsupported unit such as "km"
+    // would otherwise be silently treated as meters by the Earth-space patch builder,
+    // producing a grid orders of magnitude too fine and hanging preprocessing. Fail fast.
+    if (!unitsSupported(size.getUnits().toString())) {
```
GondekNP, line 241: Note here for posterity - we currently have a bit of a smell (see #444) where we are performing some meter / degree checks in numerous locations, without using the existing (built for this situation) `Converter` class. Intuitively, a user using `16 km` should rightly assume, based on how other m -> km conversions happen at runtime, that the same will occur here, but the `preprocess` is incorrectly interpreting this as meters and creating an enormous grid. 

For the sake of failing fast, we introduce this check, but the real solution is to route any and all unit checks through `Converter`

PR #447, opened 2026-05-29
Fix organism here.<patch-attribute> returning year-0 value forever

GondekNP:
## Summary

Fixes a runtime bug where an organism reading a patch attribute via \`here.<name>\` from its \`.step\` handler returned the value cached at the simulation's first step, even when the patch's own attribute correctly updated each step.

**Root cause:** \`MinimalEngineBridge.DecoratingShadowIterator.next()\` minted a fresh \`ShadowingEntity\` wrapper for the same underlying patch on every iteration of \`getCurrentPatches()\`. Each substep (\`init\`, \`start\`, \`step\`, \`end\`) iterates the patches separately, so \`init\` produced wrapper A and \`step\` produced wrapper B — both wrapping the same inner \`DirectLockMutableEntity\`. Organisms created during \`init\` captured \`here = A\` and held that reference forever. The simulation stepper then ran subsequent substeps against B/C/D… whose \`resolvedCacheByIndex\` was correctly refreshed each step, while A's cache permanently held its init-substep values. When an organism's \`.step\` read \`here.<patch-attribute>\`, it hit A's stale cache and returned the year-0 value.

**Fix:** Cache wrappers per patch identity in the bridge so the same \`ShadowingEntity\` instance is returned for a given inner patch across all substeps and steps. Organism \`here\` references then remain valid, and substep-boundary cache invalidation in \`startSubstep\` does the right thing for organism-visible reads.

**How it surfaced:** Reported by the josh-llm-experiment team; only manifests when an attribute has both \`.init\` and \`.step\` handlers and an organism reads it via \`here.<name>\`. The existing examples and tests read external data directly inside organisms (\`external X\`), which bypasses the patch-attribute cache path — that's why nothing in the suite caught it.

## Test plan

- [x] New JUnit reproducer \`testOrganismSeesCurrentPatchAttributeViaHere\` plus Josh fixture; fails before the fix at year 3, passes after.
- [x] \`./gradlew test\` — full suite green.
- [ ] (Recommended) josh-llm-experiment team rebuilds against a jar containing this fix and re-runs the headline batch; expect β → 1.0 / R² → 1.0 on the Josh arm to match the Mesa control.

### DIFF
--- a/examples/test/test_organism_created_midsim_sees_here.josh
+++ b/examples/test/test_organism_created_midsim_sees_here.josh
@@ -1,0 +1,54 @@
+# Test: Organism Created Mid-Simulation Sees Current here.<patch-attribute>
+#
+# Companion to test_organism_sees_stale_here_patch_attribute.josh. That
+# fixture creates the organism at init (step 0); this one creates it
+# mid-simulation (step 2) to exercise the EntityFastForwarder path:
+# a freshly created organism must be fast-forwarded so that here.<attr>
+# reads the patch's CURRENT value, not a year-0 / stale value.
+#
+# Expected behavior: from the step the organism is created onward, its
+# seenValue mirrors the patch's patchValue every step. This validates both
+# fast-forward at creation (step 2) and wrapper-identity persistence across
+# the subsequent steps (3, 4, 5) where the organism carries forward via prior.
+
+start simulation Main
+
+  grid.size = 10 m
+  grid.low.x = 0 m
+  grid.low.y = 0 m
+  grid.high.x = 30 m
+  grid.high.y = 30 m
+
+  steps.low = 0 count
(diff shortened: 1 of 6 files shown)


GondekNP:

## Independent verification of this fix against the reporter's reproducer

Claude ran the `josh-llm-experiment` 's bundled reproducer (16km ForeverTree, 63 patches × 100 yrs) against this branch both with and without the one-line change, computing β/α/R² from the output CSVs.

| build | β | α | R² | worst residual cells |
|---|---|---|---|---|
| without fix | 1.0641 | −1.66 | 0.890 | `4.5_1.5` = −40.66 m, `4.5_4.5` = +51.66 m |
| **with fix** | **1.0003** | −0.002 | **1.0000** | ±0.2 m |

